PR #2, opened 2022-01-13
Implemented code for static UI of preorder jobs screen (#1ykzq28)

azkyakhan:


src/views/PreOrder.vue
```diff
@@ -0,0 +1,124 @@
...
+    IonPage,
+    IonTitle,
+    IonToggle,
+    IonToolbar
+  },
+});
+</script>
+
+<style scoped>
+.content {
```
dt2patel, line 118: I don't think we need a class for this. we can directly tag main component

dt2patel (changes requested):
Class change.

src/views/PreOrder.vue
azkyakhan: Done sir

dt2patel (approved):
Looks good

PR #6, opened 2022-01-18
Implemented code for static UI of product sync (#22j85tf)

azkyakhan:


azkyakhan:
product

src/views/Product.vue
```diff
@@ -0,0 +1,92 @@
...
+            <ion-label>{{ $t("Sync products") }}</ion-label>
+            <ion-select :interface-options="customPopoverOptions" interface="popover">
+              <ion-select-option>Hourly</ion-select-option>
+              <ion-select-option>Every 6 hours</ion-select-option>
+              <ion-select-option>Daily</ion-select-option>
+            </ion-select>
+          </ion-item>
+           <ion-item>
+            <ion-label>{{ $t("Sync collections") }}</ion-label>
+            <ion-select :interface-options="customPopoverOptions" interface="popover">
```
dt2patel, line 25: For now lets use all popovers as components instead of inline incase we can find a way to make them reusable.

src/views/Product.vue
```diff
@@ -0,0 +1,92 @@
...
+    IonLabel,
+    IonPage,
+    IonSelect,
+    IonSelectOption,
+    IonTitle,
+    IonToolbar
+  },
+  setup() {
+    const customPopoverOptions: any = {
+    header: 'Title',
```
dt2patel, line 75: Add actual title

dt2patel (changes requested):
Please make suggested changes

src/views/Product.vue
azkyakhan: Ok sir

src/views/Product.vue
azkyakhan: Done

PR #8, opened 2022-01-18
Implemented code for static UI of inventory sync page (#22j85vj)

azkyakhan:


azkyakhan:
Inventory

src/views/Inventory.vue
```diff
@@ -0,0 +1,161 @@
...
+    IonRadio,
+    IonSelect,
+    IonSelectOption,
+    IonTitle,
+    IonToggle,
+    IonToolbar,
+  },
+  setup() {
+    const customPopoverOptions: any = {
+    header: 'Title',
```
dt2patel, line 138: Add title

src/views/Inventory.vue
```diff
@@ -0,0 +1,161 @@
...
+    }  
+  }
+});
+</script>
+
+<style scoped>
+main {
+  display: flex;  
+  align-items: flex-start;
+  max-width: 742px;
```
dt2patel, line 152: I don't think you need to define width here. There is a way to do this without defining explicit width.

dt2patel (changes requested):
Make suggested changes

src/views/Inventory.vue
azkyakhan: Done

src/views/Inventory.vue
azkyakhan: Done

PR #62, opened 2022-03-30
Implemented logic to show  alert when there are changes in the detail component(#245eu5u)

disha1202:
### Related Issues

Closes #


adityasharma7:
@disha1202 
I think we could simplify it by using lifecycle hook of the component 
https://ionicframework.com/docs/vue/lifecycle
https://vuejs.org/guide/essentials/lifecycle.html

PR #64, opened 2022-03-30
Fix: error of infinite scroll when there are no records(#24w686f)

disha1202:
### Related Issues

Closes #


adityasharma7:
@disha1202 
I think we should fix 
https://github.com/hotwax/job-manager/blob/main/src/store/modules/job/actions.ts#L120
https://github.com/hotwax/job-manager/blob/main/src/store/modules/job/actions.ts#L124

We should pass an object with empty list and total 0
https://github.com/hotwax/job-manager/blob/main/src/store/modules/job/actions.ts#L108

Same applies to history

PR #75, opened 2022-04-01
Implemented logic to store draft jobs on login and clear on logout(#24gtw1w)

disha1202:
### Related Issues

Closes #


adityasharma7 (approved):

src/store/modules/job/actions.ts
```diff
@@ -9,6 +9,39 @@ import { DateTime } from 'luxon';
 
 const actions: ActionTree<JobState, RootState> = {
 
+  async fetchDraftJobs({commit, dispatch, state}, payload){
```
adityasharma7, line 12: It seems we have not used state variable, we could remove it

adityasharma7 (changes requested):

dt2patel:
@k2maan check that we haven't already implemented this, if we haven't, review this and see if we can use this.

PR #101, opened 2022-04-22
Implemented: functionality to add refresh button when we do not have jobs on pipeline page (#28940kn).

meet-aniket:
**Screen shot:**
image

### DIFF
--- a/src/views/Pipeline.vue
+++ b/src/views/Pipeline.vue
@@ -27,7 +27,10 @@
           <!-- Empty state -->
           <div v-if="pendingJobs?.length === 0">
             <p class="ion-text-center">{{ $t("There are no jobs pending right now")}}</p>
-            <ion-button fill="outline" @click="refreshJobs()">{{ $t('retry') }}</ion-button>
+            <ion-button fill="outline" @click="refreshJobs()">
+              {{ $t('retry') }}
+              <ion-spinner v-if="isLoading" name="crescent" />
+            </ion-button>
           </div>
 
           <div v-else>
@@ -81,7 +84,10 @@
           <!-- Empty state -->
           <div v-if="runningJobs?.length === 0">
             <p class="ion-text-center">{{ $t("There are no jobs running right now")}}</p>
-            <ion-button fill="outline" @click="refreshJobs()">{{ $t('retry') }}</ion-button>
+            <ion-button fill="outline" @click="refreshJobs()">
+              {{ $t('retry') }}
+              <ion-spinner v-if="isLoading" name="crescent" />
+            </ion-button>
           </div>
 
           <div v-else>
@@ -128,7 +134,10 @@
           <!-- Empty state -->
           <div v-if="jobHistory?.length === 0">
             <p class="ion-text-center">{{ $t("No jobs have run yet")}}</p>
-            <ion-button fill="outline" @click="refreshJobs()">{{ $t('retry') }}</ion-button>
+            <ion-button fill="outline" @click="refreshJobs()">
+              {{ $t('retry') }}
+              <ion-spinner v-if="isLoading" name="crescent" />
+            </ion-button>
           </div>
 
           <div v-else>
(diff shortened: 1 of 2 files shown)


src/views/Pipeline.vue
```diff
@@ -176,6 +179,7 @@
           <JobConfiguration :title="title" :job="currentJob" :status="currentJobStatus" :type="freqType" :key="currentJob"/>
         </aside>
       </main>
+      <ion-loading v-if="isLoading" show-backdrop="true" spinner="crescent" />
```
ymaheshwari1, line 182: We will not use the ion-loading component, instead, we will use the predefined emitter to display the loader on click and dismiss the loader once the API call is a success or failure.

ymaheshwari1 (changes requested):

src/views/Pipeline.vue
```diff
@@ -311,6 +318,18 @@ export default defineComponent({
         this.store.dispatch('job/fetchJobHistory', {eComStoreId: this.getCurrentEComStore.productStoreId, viewSize:process.env.VUE_APP_VIEW_SIZE, viewIndex:0}).then(() => { event.target.complete() });
       }
     },
+    async reFetchJobs() {
```
adityasharma7, line 321: Use same method we use for pull to refresh, do not pass event and handle case

src/views/Pipeline.vue
adityasharma7, line 182: I think this is not need we should add loader the same way we do for other instances

adityasharma7 (changes requested):

src/views/Pipeline.vue
meet-aniket: Done Sir.

src/views/Pipeline.vue
meet-aniket: Done Sir

src/views/Pipeline.vue
meet-aniket: Done Sir

src/views/Pipeline.vue
```diff
@@ -262,7 +276,8 @@ export default defineComponent({
       currentJobStatus: '',
       freqType: '' as any,
       isJobDetailAnimationCompleted: false,
-      isDesktop: isPlatform('desktop')
+      isDesktop: isPlatform('desktop'),
+      isLoading: false
```
adityasharma7, line 280:
```suggestion
      isRetrying: false
```

adityasharma7 (changes requested):

src/views/Pipeline.vue
```diff
@@ -318,12 +333,22 @@ export default defineComponent({
       })
     },
     async refreshJobs(event: any) {
+      this.isRetrying = true;
       if(this.segmentSelected === 'pending') {
-        this.store.dispatch('job/fetchPendingJobs', {eComStoreId: this.getCurrentEComStore.productStoreId, viewSize:process.env.VUE_APP_VIEW_SIZE, viewIndex:0}).then(() => { event.target.complete() });
+        this.store.dispatch('job/fetchPendingJobs', {eComStoreId: this.getCurrentEComStore.productStoreId, viewSize:process.env.VUE_APP_VIEW_SIZE, viewIndex:0}).then(() => {
+          if(event) event.target.complete();
+          this.isRetrying = false;
```
ymaheshwari1, line 340: We have set `isRetrying` to false, but what if the action throws some error.

ymaheshwari1 (changes requested):

src/views/Pipeline.vue
meet-aniket: Sir, we do not return any kind of response from the action so according to me, we will always execute `.then()` statement WDYT.

dt2patel:
@azkyakhan please update UI on this to center button properly

azkyakhan:
> @azkyakhan please update UI on this to center button properly

I have updated UI on this to center retry button.
https://github.com/hotwax/job-manager/pull/101/commits/c1a625ae74ef8a19334d7b4608e38394f6537749

retry-button

adityasharma7 (approved):

PR #111, opened 2022-04-25
Implemented: functionality to copy job-related information on the clipboard (#28meb2r).

meet-aniket:
### Related Issues

Closes #

Implemented functionality to copy job-related information on the clipboard and that information is like JobId, JobName, JobDescription.

image


dt2patel (changes requested):

src/views/Pipeline.vue
```diff
@@ -62,7 +62,7 @@
                   <ion-button color="danger" fill="clear" @click.stop="cancelJob(job)">{{ $t("Cancel") }}</ion-button>
                 </div>
                 <div>
-                  <ion-button fill="clear" color="medium" slot="end">
+                  <ion-button fill="clear" color="medium" slot="end" @click.stop="copyToClipboard(job)">
```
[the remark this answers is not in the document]
azkyakhan: I think I have missed that slot by mistake, now I have removed it sir

src/views/Pipeline.vue
```diff
@@ -277,6 +279,16 @@ export default defineComponent({
     })
   },
   methods: {
+    async copyToClipboard(job: any) {
```
adityasharma7, line 282: Either pass Job information as string or improve name. We could use same method at different places, if we make passing job compulsory we wouldn't be able to reuse.

adityasharma7 (changes requested):

src/views/Pipeline.vue
meet-aniket: Sir, I updated the name of the function, because we want to pass whole as param instead of jobInformation as a string.

adityasharma7 (approved):

PR #163, opened 2022-05-23
Stop selection of job to filter while unpinning jobs

rathoreprashant:
### Related Issues
Stop selection of job to filter while unpin jobs

### DIFF
--- a/src/views/Pipeline.vue
+++ b/src/views/Pipeline.vue
@@ -367,7 +367,8 @@ export default defineComponent({
     },
     updateSelectedPinnedJob(jobEnumId: any) {
       const index = (this as any).selectedPinnedJobs.indexOf(jobEnumId);
-      if (index != -1) {
+      const pinnedJobs = new Set(this.getPinnedJobs);
+      if (index != -1 || !pinnedJobs.has(jobEnumId)) {
         (this as any).selectedPinnedJobs.splice(index, 1);
       } else {
         (this as any).selectedPinnedJobs.push(jobEnumId)


src/views/Pipeline.vue
```diff
@@ -367,7 +367,8 @@ export default defineComponent({
     },
     updateSelectedPinnedJob(jobEnumId: any) {
       const index = (this as any).selectedPinnedJobs.indexOf(jobEnumId);
-      if (index != -1) {
+      const pinnedJobs = new Set(this.getPinnedJobs);
```
adityasharma7, line 370: I think we could directly search in list

src/views/Pipeline.vue
```diff
@@ -367,7 +367,8 @@ export default defineComponent({
...
+      if (index != -1 || !pinnedJobs.has(jobEnumId)) {
```
adityasharma7, line 371: It will have wrong result in case pinnedJobs is unpinned without using in filter

adityasharma7 (changes requested):

src/views/Pipeline.vue
```diff
@@ -560,6 +559,14 @@ export default defineComponent({
     this.getPendingJobs();
     this.store.dispatch('user/getPinnedJobs');
   },
+  mounted(){
+    emitter.on("unSelectPinJob", (enumId)=> {
```
adityasharma7, line 563: Try handling it on popover close

adityasharma7 (changes requested):

src/views/Pipeline.vue
rathoreprashant: Improved sir

src/views/Pipeline.vue
```diff
@@ -502,7 +503,11 @@ export default defineComponent({
         event: ev,
         componentProps: { job }
       });
-      return popover.present()
+      emitter.on("unSelectPinJob", (enumId)=> {
+        (this as any).updateSelectedPinnedJob(enumId);
+      });
+      emitter.off("unSelectPinJob", (this as any).updateSelectedPinnedJob);
+        return popover.present()
```
adityasharma7, line 510: Indentation issues

src/views/Pipeline.vue
adityasharma7: This still needs update

src/views/Pipeline.vue
```diff
@@ -502,7 +503,11 @@ export default defineComponent({
         event: ev,
         componentProps: { job }
       });
-      return popover.present()
+      emitter.on("unSelectPinJob", (enumId)=> {
+        (this as any).updateSelectedPinnedJob(enumId);
+      });
+      emitter.off("unSelectPinJob", (this as any).updateSelectedPinnedJob);
```
adityasharma7, line 509: I this is not the right way to unsubscribe

src/views/Pipeline.vue
```diff
@@ -502,7 +503,11 @@ export default defineComponent({
         event: ev,
         componentProps: { job }
       });
-      return popover.present()
+      emitter.on("unSelectPinJob", (enumId)=> {
```
adityasharma7, line 506: We should pass the function reference as it will be needed for unsubscribe

src/views/Pipeline.vue
```diff
@@ -546,10 +551,8 @@ export default defineComponent({
     },
     async updatePinnedJobs(enumId: any) {
       const pinnedJobs = new Set(this.getPinnedJobs);
-      if(pinnedJobs.has(enumId)) {
+      if (pinnedJobs.has(enumId)) {
         pinnedJobs.delete(enumId);
-      } else {
```
adityasharma7, line 551: Why this is removed?

src/components/JobActionsPopover.vue
```diff
@@ -81,6 +82,8 @@ export default defineComponent({
       const pinnedJobs = new Set(this.getPinnedJobs);
       if(pinnedJobs.has(enumId)) {
         pinnedJobs.delete(enumId);
+        await this.store.dispatch('user/updatePinnedJobs', { pinnedJobs: [...pinnedJobs] });
+        emitter.emit("unSelectPinJob", enumId);
```
adityasharma7, line 86: Can we check if we could handle on popover dismiss

adityasharma7 (changes requested):

src/views/Pipeline.vue
rathoreprashant: this was removed because of the pipeline.vue page has not support to pin job this Method(updatePinnedJobs) is only used for unpin jobs

src/views/Pipeline.vue
rathoreprashant: Improved sir

PR #186, opened 2022-06-07
Implemented: Apply Job type on inventory page(#2ewjybr)

shashwatbangar:
### Related Issues

Closes #

### DIFF
--- a/src/store/modules/job/actions.ts
+++ b/src/store/modules/job/actions.ts
@@ -93,23 +93,24 @@ const actions: ActionTree<JobState, RootState> = {
             return { 
               id: response.jobId,
               name: response.jobName,
-              systemJobEnum: [
-                {id: response.enumId},
-                {name: response.enumName},
-                {description: response.description},
-              ],
+              systemJobEnum: {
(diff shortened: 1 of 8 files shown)


src/store/modules/job/actions.ts
```diff
@@ -437,6 +437,32 @@ const actions: ActionTree<JobState, RootState> = {
       const tempExpr = Object.values(cached).map((job: any) => job.tempExprId)
       await dispatch('fetchTemporalExpression', tempExpr)
 
+      cached = Object.values(cached).map((cachedJob: any) => ({
```
ymaheshwari1, line 440: I think this will convert the cached into an array and thus we will not be able to fetch the job data from the cached state.

ymaheshwari1 (changes requested):

src/store/modules/job/actions.ts
shashwatbangar: Right sir! I am fixing the logic.

dt2patel:
old pr, closing cause feature is implemented

PR #212, opened 2022-08-03
Implemented: additional job actions, to job config component (#364tt29)

rathoreprashant:
### Related Issues

Closes #

addititionalAction

runnow

https://user-images.githubusercontent.com/33576503/182549825-25357758-5707-44f5-8fc6-4579e118a324.mp4

### DIFF
--- a/src/components/JobConfiguration.vue
+++ b/src/components/JobConfiguration.vue
@@ -91,7 +91,7 @@
     <ion-item class="column" slot="end" @click="updatePinnedJobs(currentJob?.systemJobEnumId)" button>
       <ion-icon slot="start" :icon="pinOutline" />
       {{ $t("Pin job") }}
-      <ion-checkbox updatePinnedJobs :checked=checked slot="end"></ion-checkbox>
+      <ion-checkbox :checked="getPinnedJobs.includes(currentJob.systemJobEnumId)" slot="end"></ion-checkbox>
     </ion-item>
   </ion-list>
 
--- a/src/components/JobConfiguration.vue
+++ b/src/components/JobConfiguration.vue
@@ -70,37 +70,31 @@
       <ion-button expand="block" @click="saveChanges()">{{ $t("Save changes") }}</ion-button>
     </div>
   </section>
-  <ion-list class="row">
-    <ion-item class="column" slot="start" @click="viewJobHistory(currentJob)" button>
+  <div class="job-actions">
+    <ion-item slot="start" @click="viewJobHistory(currentJob)" button>
       <ion-icon slot="start" :icon="timeOutline" />
       {{ $t("History") }}
     </ion-item>
-
-    <ion-item class="column" slot="end" @click="runNow(title, currentJob)" button>
+    <ion-item slot="end" @click="runNow(title, currentJob)" button>
       <ion-icon slot="start" :icon="flashOutline" />
       {{ $t("Run now") }}
     </ion-item>
-  </ion-list>
-
-  <ion-list class="row">
-    <ion-item class="column" slot="start" @click="copyJobInformation(currentJob)" button>
+    <ion-item slot="start" @click="copyJobInformation(currentJob)" button>
       <ion-icon slot="start"  :icon="copyOutline" />
       {{ $t("Copy details") }}
     </ion-item>
-
-    <ion-item class="column" slot="end" @click="updatePinnedJobs(currentJob?.systemJobEnumId)" button>
+    <ion-item slot="end" @click="updatePinnedJobs(currentJob?.systemJobEnumId)" button>
       <ion-icon slot="start" :icon="pinOutline" />
       {{ $t("Pin job") }}
       <ion-checkbox :checked="getPinnedJobs.includes(currentJob.systemJobEnumId)" slot="end"></ion-checkbox>
     </ion-item>
-  </ion-list>
+  </div>
 
 </template>
 
 <script lang="ts">
 import { defineComponent } from "vue";
 import {
-  alertController,
   IonBadge,
   IonButton,
   IonCheckbox,
@@ -113,6 +107,7 @@ import {
   IonModal,
   IonSelect,
   IonSelectOption,
+  alertController,
   modalController,
 } from "@ionic/vue";
 import {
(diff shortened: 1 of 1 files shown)


src/components/JobConfiguration.vue
```diff
@@ -86,15 +113,20 @@ import {
   IonModal,
   IonSelect,
   IonSelectOption,
-  alertController
```
rathoreprashant, line 89: Need to remove unwanted changes

src/components/JobConfiguration.vue
```diff
@@ -70,13 +70,40 @@
       <ion-button expand="block" @click="saveChanges()">{{ $t("Save changes") }}</ion-button>
     </div>
   </section>
+  <ion-list class="row">
```
azkyakhan, line 73: I would suggest to not use 2 separate lists to keep the items, instead of this please keep all the items in single div and apply flex-wrap property to it and also set some flex basis to the items.

azkyakhan (changes requested):

src/components/JobConfiguration.vue
rathoreprashant: Improved

src/components/JobConfiguration.vue
rathoreprashant: Updated ma'am

src/components/JobConfiguration.vue
```diff
@@ -328,7 +416,19 @@ ion-list {
 
   .mobile-only {
     display: none;
-  }  
+  }
+  .job-actions {
```
azkyakhan, line 420: Improve indentation

azkyakhan (changes requested):

src/components/JobConfiguration.vue
rathoreprashant: improved ma'am

src/components/JobConfiguration.vue
```diff
@@ -328,7 +416,19 @@ ion-list {
...
+    margin-top: var(--spacer-sm);
```
azkyakhan, line 421: Please arrange the usage of css properties in correct way. Display should be written first and margin will be on last.

azkyakhan (changes requested):

src/components/JobConfiguration.vue
```diff
@@ -70,13 +70,34 @@
       <ion-button expand="block" @click="saveChanges()">{{ $t("Save changes") }}</ion-button>
     </div>
   </section>
+  <div class="job-actions">
+    <ion-item slot="start" @click="viewJobHistory(currentJob)" button>
```
dt2patel, line 74: @azkyakhan Thoughts on slot here?

src/components/JobConfiguration.vue
```diff
@@ -70,13 +70,34 @@
...
+      {{ $t("Run now") }}
+    </ion-item>
+    <ion-item slot="start" @click="copyJobInformation(currentJob)" button>
+      <ion-icon slot="start"  :icon="copyOutline" />
+      {{ $t("Copy details") }}
+    </ion-item>
+    <ion-item slot="end" @click="updatePinnedJobs(currentJob?.systemJobEnumId)" button>
+      <ion-icon slot="start" :icon="pinOutline" />
+      {{ $t("Pin job") }}
+      <ion-checkbox :checked="getPinnedJobs.includes(currentJob.systemJobEnumId)" slot="end" />
```
dt2patel, line 89: I am seeing slot in different places. Why not at the same place everytime?

src/components/JobConfiguration.vue
```diff
@@ -328,7 +416,19 @@ ion-list {
...
+    margin-top: var(--spacer-sm);
+    row-gap: var(--spacer-sm);
+    display: flex;
+    flex-wrap: wrap;
+  }
+  .job-actions > ion-item {
+    flex-basis: 50%;
+  }
+  .job-actions > ion-item > ion-checkbox{
+    margin: unset;
```
dt2patel, line 430: Why do you need this? @azkyakhan

src/components/JobConfiguration.vue
dt2patel, line 420: We already have an "actions" class. So job actions doesn't make sense here I think. "More actions" seems better.

dt2patel (changes requested):

src/components/JobConfiguration.vue
azkyakhan: This slots over here are not needed.

src/components/JobConfiguration.vue
azkyakhan: Due to the margin on checkbox this item is looking like this :
Screenshot from 2022-08-05 16-44-24

That is why we have unset the margin on checkbox.

dt2patel: I don't see problem here with checkbox spacing. Am I missing something?

azkyakhan: That item line is slightly below from the adjacent item. 
Untitled design

dt2patel: Got it, what if we instead align content/items to end. I'd like to avoid changing item height from native behavior.

azkyakhan: I tried using `align-items` property but it is working when we are providing `flex-end` value to it :
Screenshot from 2022-08-08 10-54-19


If we are giving `end` as value then their is no change :
Screenshot from 2022-08-08 10-54-46

src/components/JobConfiguration.vue
rathoreprashant: Arranged ma'am

src/components/JobConfiguration.vue
rathoreprashant: Removed

src/components/JobConfiguration.vue
rathoreprashant: Updated sir

src/components/JobConfiguration.vue
rathoreprashant: Implemented align-items to end

src/components/JobConfiguration.vue
rathoreprashant: updated sir

PR #238, opened 2022-09-09
Fixed: error in console due to empty shopify configuration on logout(#2tky1f3)

disha1202:
### Related Issues

Closes #

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -28,7 +28,7 @@
           <ion-label class="ion-text-wrap">
             <p class="overline">{{ instanceUrl }}</p>
             {{ eComStore.storeName }}
-            <p>{{ currentShopifyConfig?.name ? currentShopifyConfig.name : currentShopifyConfig?.shopifyConfigName }}</p>
+            <p>{{ currentShopifyConfig.name ? currentShopifyConfig.name : currentShopifyConfig.shopifyConfigName }}</p>
           </ion-label>
           <ion-note slot="end">{{ userProfile?.userTimeZone }}</ion-note>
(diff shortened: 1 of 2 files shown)


src/components/Menu.vue
```diff
@@ -28,7 +28,7 @@
           <ion-label class="ion-text-wrap">
             <p class="overline">{{ instanceUrl }}</p>
             {{ eComStore.storeName }}
-            <p>{{ currentShopifyConfig.name ? currentShopifyConfig.name : currentShopifyConfig.shopifyConfigName }}</p>
+            <p>{{ currentShopifyConfig?.name ? currentShopifyConfig.name : currentShopifyConfig?.shopifyConfigName }}</p>
```
ymaheshwari1, line 31: WDYT?
```suggestion
            <p v-if="currentShopifyConfig">{{ currentShopifyConfig.name ? currentShopifyConfig.name : currentShopifyConfig.shopifyConfigName }}</p>
```

ymaheshwari1 (changes requested):

src/components/Menu.vue
adityasharma7, line 31: I think we should set it as empty object by default and this check will not be required then

adityasharma7 (changes requested):

src/components/Menu.vue
disha1202: Set empty object by default to resolve this issue

disha1202: Set empty object by default to resolve this issue

adityasharma7 (approved):

PR #315, opened 2022-12-16
Fixed: Jobs getting scheduled multiple times(#32qywdd)

### LINKED ISSUE #311: Static jobs should be scheduled once
## Current behavior

Navigate to any page, order, inventory etc where the job is disabled. Try scheduling the job from job configuration section/page, after the success message, again click on save changes it will schedule another instance of the job.

## Expected behavior

Users should be allowed to schedule it once only. Save changes should be disabled till current action is not completed and once done job scheduled should be updated through it

shashwatbangar:
### Related Issues

Closes #311 

Earlier, when we navigate to any page, order, inventory etc where the job is disabled & try scheduling the job from job configuration section/page, after the success message, again click on save changes it will schedule another instance of the job. 

Reason being when we click on SAVE CHANGES again, the currentJob is not yet updated or more specifically, marked as scheduled. This was because after hitting the scheduleJob API, response was not handled correctly and set in currentJob state.

As a result when we click on SAVE CHANGES, currentJob shows that job is still in DRAFT state hence will get scheduled again whereas what should happen is, when we schedule a job for the first time, it should get updated and marked as PENDING. So when we update this job or disable it, job will get updated not get scheduled again.

https://user-images.githubusercontent.com/62360781/208084725-7b56d6a5-c337-4036-9729-0dd4bae68905.mov


src/store/modules/job/actions.ts
```diff
@@ -458,19 +458,17 @@ const actions: ActionTree<JobState, RootState> = {
...
+        const jobs = await dispatch('fetchJobs', {
           inputFields: {
             'systemJobEnumId': payload.systemJobEnumId,
             'systemJobEnumId_op': 'equals'
           }
         })
-        if(jobs.status === 200 && !hasError(jobs) && jobs.data?.docs.length) {
-          jobs = jobs.data?.docs;
-          const job = jobs.find((job: any) => job?.jobId === payload.jobId);
+        const job = jobs[1];
```
ymaheshwari1, line 467: Why we are accessing values using index? As there is no gurantee that the specific job will come on that index only.

ymaheshwari1 (changes requested):

src/store/modules/job/actions.ts
shashwatbangar: Right sir. Fixed it.

src/store/modules/job/actions.ts
```diff
@@ -458,19 +458,17 @@ const actions: ActionTree<JobState, RootState> = {
...
           inputFields: {
             'systemJobEnumId': payload.systemJobEnumId,
             'systemJobEnumId_op': 'equals'
           }
         })
-        if(jobs.status === 200 && !hasError(jobs) && jobs.data?.docs.length) {
-          jobs = jobs.data?.docs;
-          const job = jobs.find((job: any) => job?.jobId === payload.jobId);
+        const job = jobs.find((job: any) => job?.jobId === payload.jobId);
+        if(job.status === 200 && !hasError(job) && job.data?.docs.length) {
```
adityasharma7, line 468: Improve variable name job, status, data etc are attributes of response not job

adityasharma7 (changes requested):

PR #324, opened 2022-12-26
Improved Menu footer by providing option to change shopifyConfig. (#85zrgqc2a)

### LINKED ISSUE #316: Menu footer improvement
## What is the motivation for adding/enhancing this feature?

The menu footer should have the option to change product store and Shopify config when there are multiple options. By default product options should be there and if there are multiple product stores & Shopify config then only product store options should exist

For most of the cases either there will be multiple stores or the shopify configs. In any of the scenario right configuration should be there in menu footer

## What are the acceptance criteria?

- [ ] ...

## Can you complete this feature request by yourself?

- [ ] YES
- [ ] NO

k2maan:
### Related Issues

Closes #316 

The user can now change the store and the config. both from the footer, only if multiple options are available. If both can be changed, only the option to change the store is there.

Job Manager - HotWax Commerce (1)


src/components/Menu.vue
```diff
@@ -34,14 +34,26 @@
           </ion-label>
           <ion-note slot="end">{{ userProfile?.userTimeZone }}</ion-note>
         </ion-item>
-        <ion-item lines="none">
-          <ion-select interface="popover" :value="eComStore.productStoreId" @ionChange="setEComStore($event)">
+        <ion-item v-if="userProfile?.stores.length > 2" lines="none">
+          <ion-select  interface="popover" :value="eComStore.productStoreId" @ionChange="setEComStore($event)">
```
ymaheshwari1, line 38: Unwanted change.

src/components/Menu.vue
```diff
@@ -34,14 +34,26 @@
...
+          <ion-select interface="popover" :value="currentShopifyConfig?.shopifyConfigId" @ionChange="setShopifyConfig($event)">
+            <ion-select-option v-for="shopifyConfig in shopifyConfigs" :key="shopifyConfig.shopifyConfigId" :value="shopifyConfig.shopifyConfigId" >{{ shopifyConfig.name ? shopifyConfig.name : shopifyConfig.shopifyConfigName }}</ion-select-option>
+          </ion-select>
+        </ion-item>
+        <ion-item v-else lines="none">
           <ion-label class="ion-text-wrap">
-            <p>{{ currentShopifyConfig.name ? currentShopifyConfig.name : currentShopifyConfig.shopifyConfigName }}</p>
+           <p>
+             {{ currentShopifyConfig.name ? currentShopifyConfig.name : currentShopifyConfig.shopifyConfigName }}
+           </p>
```
ymaheshwari1, line 56: Unwanted change

src/components/Menu.vue
```diff
@@ -104,16 +116,22 @@ export default defineComponent({
...
   methods: {
     async setEComStore(event: CustomEvent) {
       if(this.userProfile) {
         await this.store.dispatch('user/setEcomStore', { 'productStoreId': event.detail.value })
-        emitter.emit("productStoreChanged")
+        emitter.emit("productStoreOrConfigChanged")
       }
     },
+    async setShopifyConfig(event: CustomEvent){
+      await this.store.dispatch('user/setCurrentShopifyConfig', { 'shopifyConfigId': event.detail.value });
```
ymaheshwari1, line 132: In `setEComStore` we are checking for `userProfile` but in this method  we are not, any specific reason?

ymaheshwari1 (changes requested):

src/components/Menu.vue
k2maan: The code is referenced from the Settings page. There, we are doing the same. 

I think the reason for using it on the Settings page is that the config setting is dependent on the ecomStores, hence if we check for the user profile before setting the ecomStore, we do not need to check the same for setting the configs.

Now, coming to our case, the Menu footer, on second thought, I think we do not need to check the same here, for any of the cases, as `this.userProfile` checks if the user is logged in or not and on tracing the code, only on successful login we are able to use the app and route guarding is there too.

So, I do not think it is required in any of the cases.

src/components/Menu.vue
k2maan: Reverted sir

src/components/Menu.vue
k2maan: Reverted sir

src/components/Menu.vue
k2maan: As discussed, sir, the check is not needed here since we only need to handle this case when logging out from the settings page.

disha1202 (approved):

adityasharma7 (approved):

PR #364, opened 2023-02-06
Fixed: search jobs modal showing 'no jobs found' without searching any job (#85zrmar8z)

### LINKED ISSUE #358: Bulk Job scheduling: 'No jobs found' comes when we open the select jobs link

- [ ] YES
- [ ] NO

## Environment details

- Browser: 
- OS: 
- Code Version:

k2maan:
### Related Issues

Closes #358 

The select jobs modal showed 'No jobs found' without even searching for a job. Fixed the issue and now the modal shows 'Searched jobs will appear here' if no jobs are searched. 

Job Manager - HotWax Commerce


src/views/SelectJobsModal.vue
```diff
@@ -12,7 +12,11 @@
   <ion-content>
     <ion-searchbar v-model="queryString" :placeholder="$t('Search jobs')" @keyup.enter="search($event)" />
 
-    <div v-if="jobs.length === 0" class="ion-text-center">
+    <div v-if="!isJobSearched" class="ion-text-center">
+      <p>{{ $t("Searched jobs will appear here") }}</p>
+    </div>    
+
+    <div v-else-if="jobs.length === 0" class="ion-text-center">
```
disha1202, line 19: Use `queryString` instead of `isJobSearched`

src/views/SelectJobsModal.vue
```diff
@@ -12,7 +12,11 @@
   <ion-content>
     <ion-searchbar v-model="queryString" :placeholder="$t('Search jobs')" @keyup.enter="search($event)" />
 
-    <div v-if="jobs.length === 0" class="ion-text-center">
+    <div v-if="!isJobSearched" class="ion-text-center">
```
disha1202, line 15: Also we can remove this.

disha1202 (changes requested):

src/views/SelectJobsModal.vue
k2maan: As queryString is v-modelled with the search bar, even if we type something, without searching (pressing enter) it'll show 'No jobs found' in that case.

adityasharma7: Instead of using v-model, we could update the queryString only on enter press. This way we will not need flag variable

adityasharma7 (changes requested):

src/views/SelectJobsModal.vue
```diff
@@ -92,7 +96,9 @@ export default defineComponent({
   methods: {
     async search(event: any) {
       this.queryString = event.target.value.trim();
-      if(this.queryString.length > 0) this.getJobs();
+      if(this.queryString.length > 0) {
```
adityasharma7, line 99: Why do we need the length check at all?

_"That is, all values are truthy except false, 0, -0, 0n, "", null, undefined, and NaN."_

https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/views/SelectJobsModal.vue
```diff
@@ -10,9 +10,13 @@
...
   <ion-content>
-    <ion-searchbar v-model="queryString" :placeholder="$t('Search jobs')" @keyup.enter="search($event)" />
+    <ion-searchbar :placeholder="$t('Search jobs')" @keyup.enter="search($event)" />
 
-    <div v-if="jobs.length === 0" class="ion-text-center">
+    <div v-if="queryString.length === 0" class="ion-text-center">
+      <p>{{ $t("Searched jobs will appear here") }}</p>
+    </div>    
+
+    <div v-else-if="jobs.length === 0 && queryString.length != 0" class="ion-text-center">
```
adityasharma7, line 19: I think the condition `queryString.length != 0"` might not be needed, else-if will execute only when `queryString.length === 0` fails.

adityasharma7 (changes requested):

src/views/SelectJobsModal.vue
k2maan: Yes, sir. The empty queryString itself is a false value.

src/views/SelectJobsModal.vue
k2maan: Improved.

PR #667, opened 2024-01-24
Implemented: job refresh functionality for job completed already but not updated in state and UI (#660)

### LINKED ISSUE #660: Cancelling a job still shows it in pending state
## Current behavior

When we schedule a job or open an already pending job and then cancelling the job after its runTime is passed, the toast shows that the job is updated, but still the job remains in the pending state.

## Expected behavior

Cancelling a job that has either failed or completed should result in an error message

- Go to any page and open any job card which is in pending state(for ex: on order page, open new orders job)
- Wait until the jobs runTime is not passed
- Disable the job from the card
- Observe that the toast displays a success message and the api also return success but on refresh the job is still in pending state

- [x] YES
- [ ] NO

## Environment details

- Browser: 
- OS: 
- Code Version: 

## Additional information

The observation here is that, when we open the job configuration card, the card displays information for a specific job(Ex. jobId = 111), now when the runTime for the job 111 is passed, the status of the job has been changed (Complete, fail, or queue), but as such information is not available on the app side, the configuration card still holds the old information for the job 111, now when cancelling the job 111 the api returns success and thus app displays a success messge, but as the job has already been processed before cancelling so a new job with another id 112 is available and gets displayed. This results in a behaviour which feels like the job is not getting cancelled.

amansinghbais:
### Related Issues

Closes #660

Currently, when we schedule a job or open an already pending job and then cancelling the job after its runTime is passed, the toast shows that the job is updated, but still the job remains in the pending state.

Hence update the logic of JobConfiguration component and pipeline page for handling these cases as follows: 
JobConfiguration component
- Implemented logic to disable some actions when user try performing action on job that has already completed but still showing in UI.
- Shows toast on action post job completion as shown below:
- Implemented a refresh button in header of card which allow users to fetch updated details of current job.

Pipeline page
- Shows toast on skip or cancel actions on jobs which are already completed and past its runTime.
- Refetch all jobs, clear currentJob from state.

Job Configuration Component
Screenshot from 2024-01-24 14-56-40

Screenshot from 2024-01-24 10-58-25

Pipeline page
Screenshot from 2024-01-24 14-57-00


src/components/JobConfiguration.vue
```diff
@@ -3,7 +3,10 @@
     <ion-item lines="none">
       <!-- Adding conditional check for currentJob.jobName as currentJob is undefined when i18n runs $t -->
       <h1>{{ isBrokerJob ? currentJob.jobName : currentJob.enumName ? currentJob.enumName : currentJob.jobName ? currentJob.jobName : '' }}</h1>
-      <ion-badge slot="end" color="dark" v-if="currentJob?.runTime && currentJob.statusId !== 'SERVICE_DRAFT'">{{ $t("running") }} {{ timeTillJob(currentJob.runTime) }}</ion-badge>
+      <ion-button fill="outline" slot="end" v-if="isRefreshEnabled" @click="refreshCurrentJob">
```
ymaheshwari1, line 6:
```suggestion
      <ion-button fill="outline" slot="end" v-if="isRefreshRequired" @click="refreshCurrentJob">
```

src/components/JobConfiguration.vue
```diff
@@ -3,7 +3,10 @@
...
+        <ion-icon :icon="refreshOutline" />
```
ymaheshwari1, line 7:
```suggestion
        <ion-icon :icon="refreshOutline" slot="icon-only"/>
```

src/components/JobConfiguration.vue
```diff
@@ -246,6 +252,27 @@ export default defineComponent({
       this.runTime = currentRunTime
       this.runTimes = runTimes
     },
+    isJobPassed() {
```
ymaheshwari1, line 255:
```suggestion
    isRuntimePassed() {
```

src/components/JobConfiguration.vue
```diff
@@ -246,6 +252,27 @@ export default defineComponent({
...
+      return this.currentJob.runTime <= DateTime.now().toMillis()
+    },
+    async refreshCurrentJob() {
+      let job;
+
+      if(this.$route.path === '/pipeline') {
+        job = this.pendingJobs.find((job: any) => job.systemJobEnumId === this.currentJob.systemJobEnumId)
+      } else if(this.$route.path === '/brokering') {
+        job = this.getJob(this.currentJob.systemJobEnumId)?.find((job: any) => job.parentJobId === this.currentJob.parentJobId)
```
ymaheshwari1, line 264: Add a comment about the check on `parentJobId`

src/views/Pipeline.vue
```diff
@@ -499,6 +502,13 @@ export default defineComponent({
             {
               text: this.$t('Skip'),
               handler: async () => {
+                if(this.isJobPassed(job)) {
+                  await this.refreshJobs(undefined, true)
+                  showToast(this.$t("Job runtime has passed. The job data has refreshed. Please try now."))
```
ymaheshwari1, line 507:
```suggestion
                  showToast(this.$t("Job runtime has passed. The job data has refreshed. Please try again."))
```

ymaheshwari1 (changes requested):

ymaheshwari1 (changes requested):
Also check on other configuration components, like InitialJobConfiguration and fix this case.

Issue #714, opened 2024-07-01
Fix the positioning of the radio items label in the customFrequency modal.


amansinghbais:
## Current behavior

Both the lable and the radio button is coming at the start of the item which is looking distorted.

## Expected behavior

Fix the positioning of the ion-radio items label and the circleOutline.

## Screenshot
![Screenshot from 2024-07-01 15-03-07](https://github.com/hotwax/job-manager/assets/69574321/4f6a119e-b7c3-4c0e-b3cb-73314f2767b1)

## Steps to reproduce the issue

1. Open job-configuration of any job in Pipeline page.
2. Select the custom option in frequency selector. Custom frequency modal is open up.
3. You can see that both radio label and circle outline are coming at the start of the item.

## Can you handle fixing this bug by yourself?

- [ ] YES
- [ ] NO

## Environment details

- Browser: 
- OS: 
- Code Version:

PR #716, opened 2024-07-11
Fixed: made changes in the frequency code , such that both label and radio button are in sorted position

Ridwan6947:
### Related Issues

#714 

Ensured the ion-radio and ion-label elements are within the ion-item element. The ion-radio element is placed before the ion-label element and assigned the slot="start" attribute to position it at the start of the ion-item.

Screenshot from 2024-07-11 19-46-19

### DIFF
--- a/src/components/CustomFrequencyModal.vue
+++ b/src/components/CustomFrequencyModal.vue
@@ -20,8 +20,10 @@
     <div v-else>
       <ion-list>
         <ion-radio-group v-model="frequencyId">
+          <!-- Use ion-label to place the label after the radio button -->
           <ion-item :key="customFrequency.tempExprId" v-for="customFrequency in customFrequencies">
-            <ion-radio :value="customFrequency.tempExprId" slot="start">{{ customFrequency.description }}</ion-radio>
+            <ion-radio :value="customFrequency.tempExprId" slot="start"></ion-radio>
+            <ion-label class="radio-label">{{ customFrequency.description }}</ion-label>
           </ion-item>
         </ion-radio-group>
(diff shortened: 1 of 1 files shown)


src/components/CustomFrequencyModal.vue
```diff
@@ -20,8 +20,10 @@
     <div v-else>
       <ion-list>
         <ion-radio-group v-model="frequencyId">
+          <!-- Use ion-label to place the label after the radio button -->
           <ion-item :key="customFrequency.tempExprId" v-for="customFrequency in customFrequencies">
-            <ion-radio :value="customFrequency.tempExprId" slot="start">{{ customFrequency.description }}</ion-radio>
+            <ion-radio :value="customFrequency.tempExprId" slot="start"></ion-radio>
+            <ion-label class="radio-label">{{ customFrequency.description }}</ion-label>
```
ymaheshwari1, line 26: Using ion-label separately will give warnings and its not a good practice. Explore on some other way to handle this.

ymaheshwari1 (changes requested):

src/components/CustomFrequencyModal.vue
```diff
@@ -20,8 +20,9 @@
     <div v-else>
       <ion-list>
         <ion-radio-group v-model="frequencyId">
+          <!-- Use ion-label to place the label after the radio button -->
```
ymaheshwari1, line 23: Is this comment relevant? as we have not used ion-label.

ymaheshwari1 (changes requested):

Issue #680, opened 2024-02-21
Display the associated Data Manager Log ID with the failed job on the Job Details page.


AjinkyaM1:
## What is the motivation for adding/enhancing this feature?
Enhance user visibility by displaying the associated Data Manager Log ID alongside failed job details on the Job Details page, improving troubleshooting efficiency.

## What are the acceptance criteria?

- [ ] ...

## Can you complete this feature request by yourself?

- [ ] YES
- [ ] NO

dt2patel:
https://docs.google.com/document/d/1oNuOPSQafWHLvXCQZ145vZdZtEUB3wMaXUx37OruVW0/edit?usp=sharing

UI doc for this feature

dt2patel:
Badge color needs to be as follows:
<img width="307" alt="Screenshot 2024-11-14 at 3 37 53 PM" src="https://github.com/user-attachments/assets/5b3684b8-18ca-4074-9694-93fb6095c2fb">

Adameddy26:
Found one issue  - On clicking the 'View Details' button, the Import Logs page opens as expected. However, when changing the product store, the Import Logs page remains open.

AjinkyaM1:
Found some scenarios, listed below:

1. The Data Manager component becomes fully visible only when the page is scrolled down to the bottom. Attached is the Jam for more clarity.
2. MOD_ORD_ATTR: For this job, there are 678 Data Manager logs in the ADOC SV UAT instance, but the app is only displaying 250.

Jam: https://jam.dev/c/ec3d24cd-06a4-4324-899f-41870bb6e327

AjinkyaM1:
@dt2patel 
I’ve found it challenging to locate the Data Manager logs for specific jobs, especially when errors occur. For instance, if an order import job is scheduled to run every 15 minutes and generates errors, by the time the user notices, the job might have run 10-20 times. This makes it hard to find the logs related to the failed runs. 

Is it possible to add a filter that only shows jobs with errors ?

Let me know your thoughts

dt2patel:
1. Good idea. Will need to do dev beyond scope of this ticket.
2. Yes we have a limit of 250 on the view right now. Will expand later if needed but I think realistically this is fine for now

PR #729, opened 2024-09-20
Implemented: Updated the DataManager Logs page to be dynamic in Job Manager(#680)

R-Sourabh:
### Related Issues

#680 

- Added a dynamic section where details related to the files used in jobs are displayed below the main job information.
- Implemented a detail page that allows users to view the Data Manager overview from the previous page, along with other MDM configuration details on the right-hand side.
- Included a dynamic section where individual Data Manager logs are listed.
- Added ion-chips to dynamically update the list according to the job status.
- Added a popover option to download file.

Screenshot from 2024-09-20 20-19-55
Screenshot from 2024-09-20 20-19-51


src/components/DownloadLogsFilePopover.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <ion-content>
+    <ion-list>
+      <ion-list-header>{{ translate('Log Id') }}</ion-list-header>
```
ravilodhi, line 4: This will the data manager log ID not the text "Log Id".

src/components/DownloadLogsFilePopover.vue
```diff
@@ -0,0 +1,85 @@
...
+      <ion-item button @click="downloadLogFile('logFile')">
+        {{ translate('Log file') }}
+      </ion-item>
+      <ion-item button :lines="log?.errorRecordContentId ? undefined : 'none'" @click="downloadLogFile('uploadedFle')">
```
ravilodhi, line 8: Correct typo, uploadedFle.

ravilodhi, line 8: I think we should show error item disabled instead of hiding, if there is not error content.

src/components/DownloadLogsFilePopover.vue
```diff
@@ -0,0 +1,85 @@
...
+    IonItem,
+    IonList,
+    IonListHeader
+  },
+  computed: {
+    ...mapGetters({
+      getDataResourceId: 'job/getDataResourceIds',
+    }),
+  },  
+  props: ["log"],
```
ravilodhi, line 46: Rename this to dataManagerLog.

src/components/DownloadLogsFilePopover.vue
```diff
@@ -0,0 +1,85 @@
...
+  methods: {
+    async downloadLogFile(type: any) {
+      let contentIdType;
+      let dataResource = {} as any;
+
+      if (type === 'logFile') {
+        contentIdType = 'logFileContentId';
```
ravilodhi, line 53: Rename contentIdType to contentId and fetch the contentId in the if/else block itself.

src/store/modules/job/index.ts
```diff
@@ -42,6 +42,10 @@ const jobModule: Module<JobState, RootState> = {
         runtime: '',
         frequency: '',
       },
+      logs: {
```
ravilodhi, line 45: Rename logs to dataManagerLogs

src/store/modules/job/getters.ts
```diff
@@ -85,6 +85,12 @@ const getters: GetterTree <JobState, RootState> = {
     getGlobalFreq: (state) => {
       return state.bulk.frequency;
     },
+    getDataLogs: (state) => {
+      return state.logs.list
+    },
+    getDataResourceIds: (state) => (contentId: string) => {
```
ravilodhi, line 91: This should be getDataResourceId

src/store/modules/job/actions.ts
```diff
@@ -260,6 +260,96 @@ const actions: ActionTree<JobState, RootState> = {
       showToast(translate("Something went wrong"));
     })
   },
+
+  async fetchDataManagerLogs ({ commit, dispatch }, params) {
```
ravilodhi, line 264: Rename params to configId

src/store/modules/job/actions.ts
```diff
@@ -260,6 +260,96 @@ const actions: ActionTree<JobState, RootState> = {
...
+    commit(types.JOB_DATA_MANAGER_LOGS_UPDATED, []);
+    let logs = [] as any
+    const payload = {
+      "inputFields":  {
+        "statusId": ["SERVICE_CANCELLED", "SERVICE_CRASHED", "SERVICE_FAILED", "SERVICE_FINISHED", "SERVICE_PENDING", "SERVICE_RUNNING", "SERVICE_QUEUED"],
```
ravilodhi, line 269: No need to apply any filter on status.

src/store/modules/job/actions.ts
```diff
@@ -260,6 +260,96 @@ const actions: ActionTree<JobState, RootState> = {
...
+              name: contentId.coContentName
+            };
+          }
+          return acc;
+        }, {});
+      }
+    }).catch((err: any) => {
+      logger.error(err);
+    })
+    commit(types.JOB_DATA_RESOURCE_IDS_UPDATED, dataResourceIds);
```
ravilodhi, line 325: Instead of storing dataResourceIds separately, I think you should add the needed field in the data manager config record itself.

src/store/modules/job/actions.ts
```diff
@@ -260,6 +260,96 @@ const actions: ActionTree<JobState, RootState> = {
...
+  },
+  
+  async fetchDataManagerConfig({ commit }, params) {
```
ravilodhi, line 328: As we are not storing anything to state here, we can skip using action here. Service itself should fulfil the purpose.

src/store/modules/job/JobState.ts
```diff
@@ -33,4 +33,8 @@ export default interface JobState {
       runtime: any,
       frequency: any,
     },
+    logs: {
```
ravilodhi, line 36: Rename this to dataManagerLogs

src/components/DownloadLogsFilePopover.vue
```diff
@@ -0,0 +1,85 @@
...
+      } else if (type === 'failedRecords') {
+        contentIdType = 'errorRecordContentId';
+      } else if (type === 'uploadedFle') {
```
ravilodhi, line 56: Fix typo uploadedFle.

src/components/DownloadLogsFilePopover.vue
```diff
@@ -0,0 +1,85 @@
...
+        dataResource.name = this.log.contentName
+      } 
+
+      if (contentIdType) {
+        dataResource = this.getDataResourceId(this.log[contentIdType]);
+      }
+
+      if (dataResource) {
+        try {
+          const response = await JobService.downloadCsv({
```
ravilodhi, line 67: Rename this service name to something like fetchFileData.

src/utils/index.ts
```diff
@@ -341,6 +341,21 @@ const copyToClipboard = async (value: string, text?: string) => {
   });
 }
 
+const responseFileType = (response: any, fileName: string) => {
```
ravilodhi, line 344: We can merge these two functions. responseFileType anyways is not generating any meaning. You can name it like saveDataFile.

src/components/ImportLogsDetail.vue
```diff
@@ -0,0 +1,281 @@
...
+              <ion-icon slot="start" :icon="optionsOutline" />
+              {{ translate('Mode') }}
+              <ion-label slot="end">{{ configDetails?.executionModeId ? configDetails?.executionModeId : "-"}}</ion-label>
+            </ion-item>
+          </ion-list>
+        </div>
+      </div>
+
+      <div class="ion-padding">
+        <ion-chip v-for="(chip, index) in chips" :key="index" outline @click="updateLogs(chip.label)">
```
ravilodhi, line 67: The list name should be logFilters or dataManagerLogFilters instead of chips.

ravilodhi, line 67: The updateLogs

src/components/ImportLogsDetail.vue
```diff
@@ -0,0 +1,281 @@
...
+            {{ getDateTime(log.startDateTime) }}
+            <p>{{ translate('Started') }}</p>
+          </ion-label>
+  
+          <ion-label>
+            {{ getDateTime(log.finishDateTime) }}
+            <p>{{ translate('Finished') }}</p>
+          </ion-label>
+          
+          <ion-badge color="success" v-if="log.statusId === 'SERVICE_FINISHED'">{{ translate('Finished') }}</ion-badge>
```
ravilodhi, line 97: Display the status description here. You need to fetch the status data for statusTypeId='SERVICE_STATUS' and then create uiLables for all the statuses.

src/components/ImportLogsDetail.vue
```diff
@@ -0,0 +1,281 @@
...
+    IonPage,
+    IonSpinner,
+    IonTitle,
+    IonToolbar,
+  },
+  data() {
+    return {
+      configDetails: {},
+      selectedChip: 'All',
+      chips: [
```
ravilodhi, line 155: This should be renamed to logFilters and it should be the array of object, each filter object with key and value. So that when apply filters you can pass the ID.

src/components/ImportLogsDetail.vue
```diff
@@ -0,0 +1,302 @@
+<template>
```
ravilodhi, line 1: Rename this file go generic one, something like DataManagerLogDetail.vue

src/components/DataManagerLogDetails.vue
```diff
@@ -288,9 +286,10 @@ section {
 .header {
   display: grid;
   grid-template-columns: repeat(2, 1fr);
+  grid-gap: var(--spacer-sm);
 }
 
-.config {
+.config-details {
   align-self: end;
```
dt2patel, line 293: I don't think this is needed. The content should instead expand to fill the remaining space. both grid columns should take equal space. The Width of the grid should be restricted to control the width of its children.

dt2patel (changes requested):

src/components/DataManagerLogDetails.vue
R-Sourabh: The align-self property is aligning the grid column to the end vertically, not horizontally, as the layout suggests. I used this property to position the configuration name section at the bottom vertically. After removing it, this is how it looks:

Screenshot from 2024-10-18 11-06-48

dt2patel: makes sense, I mixed up justify and align. Thanks for clarifying. It was correct the way you had it

src/components/DataManagerLogDetails.vue
```diff
@@ -0,0 +1,301 @@
...
+    ...mapGetters({
+      currentJob: 'job/getCurrentJob',
+      getDataManagerLogs: 'job/getDataManagerLogs',
+      getJob: 'job/getJob',
+      getStatusDesc: 'util/getStatusDesc',
+    }),
+  },
+  async mounted() {
+    await this.fetchJobs();
+    const job = await this.getJob(this.$route.params.jobId)
```
ymaheshwari1, line 170: We can use props in place of using route.params

ymaheshwari1 (changes requested):

src/views/Pipeline.vue
```diff
@@ -608,6 +608,9 @@ export default defineComponent({
       this.freqType = appFreqType ? appFreqType[1] : "default"
 
       await this.store.dispatch('job/updateCurrentJob', { job });
+      if (job.runtimeData?.configId) {
```
ymaheshwari1, line 611: We can add a check here that we only need to fetch this logs for history jobs only.

src/views/DataManagerLogDetails.vue
```diff
@@ -0,0 +1,314 @@
...
+        resp = await JobService.fetchJobInformation({
+          "inputFields": {
+            "jobId": jobId
+          },
+          "noConditionFind": "Y"
+        })
+        if(resp.status === 200 && resp.data.docs?.length > 0 && !hasError(resp)) {
+          return resp.data.docs[0];
+        } else {
+          return [];
```
ymaheshwari1, line 219: Make consistent use of return statement, if we are returning data some section of a method then it should always return data for all the cases. In the current method we are returning data from `if` and `else` but are not returning any data from `catch` or before method execution is completed.

src/views/DataManagerLogDetails.vue
```diff
@@ -0,0 +1,314 @@
...
+  },
+  computed: {
+    ...mapGetters({
+      getDataManagerLogs: 'job/getDataManagerLogs',
+      getStatusDesc: 'util/getStatusDesc',
+    }),
+  },
+  props: ['jobId'],
+  async mounted() {
+    this.currentJob = await this.fetchJobHistory(this.jobId);
```
ymaheshwari1, line 169: What if we do not the job information?

src/components/JobConfiguration.vue
```diff
@@ -19,33 +19,45 @@
...
-          <ion-content force-overscroll="false">
-            <ion-datetime          
-              show-default-buttons
-              hour-cycle="h23"
-              :value="runTime ? (isCustomRunTime(runTime) ? getDateTime(runTime) : getDateTime(DateTime.now().toMillis() + runTime)) : getNowTimestamp()"
-              @ionChange="updateCustomTime($event)"
-            />
-          </ion-content>
-        </ion-modal>
+        <template v-if="historyJobConfig">
```
ymaheshwari1, line 22: We should not allow any edits in the job data when in history section, currently we have not added any check on editing of custom job parameter modal, it should be view only.

src/views/Pipeline.vue
```diff
@@ -180,7 +180,7 @@
           </div>
 
           <div v-else>
-          <ion-card v-for="job in jobHistory" :key="job.jobId">
+          <ion-card v-for="job in jobHistory" :key="job.jobId" @click="viewJobConfiguration(job)">
```
ymaheshwari1, line 183: We have some click events defined in child DOM of the ion-card, just cross check that those will not collide with this once.

src/components/JobConfiguration.vue
ymaheshwari1: Also update the time string to display the time till runNow, currently we have used `timeTillJob` method but that returns a string like `running in .....` but for history we need to display `..... ago`, so update the method usage for history method.

ymaheshwari1 (changes requested):

src/components/JobConfiguration.vue
```diff
@@ -430,6 +484,10 @@ export default defineComponent({
       const timeDiff = DateTime.fromMillis(time).diff(DateTime.local());
       return DateTime.local().plus(timeDiff).toRelative();
     },
+    timeFromNow (time: any) {
```
ymaheshwari1, line 487: This method is not required, as have a similar method `timeTillJob`.

src/components/JobParameterModal.vue
```diff
@@ -8,7 +8,7 @@
       </ion-buttons>
       <ion-title>{{ translate('Custom Parameters') }}</ion-title>
       <ion-buttons slot="end">
-        <ion-button color="primary" :disabled="currentJob.statusId === 'SERVICE_PENDING'" @click="save()">{{ translate('Save') }}</ion-button>
+        <ion-button color="primary" :disabled="isSaveDisabledByStatus(currentJob.statusId)" @click="save()">{{ translate('Save') }}</ion-button>
```
ymaheshwari1, line 11: We can instead revert this check to `:disabled="currentJob.statusId !== 'SERVICE_DRAFT'"`, as we only need to enable save button for draft jobs.

Issue #762, opened 2025-01-22
Import Logs not visible on short screens


dt2patel:
## Current behavior

On shorter screens the import logs cart is not visible or cutoff. We need to update the styling to allow the job configurations card to scroll independently of the pipeline section.

PR #763, opened 2025-01-22
Add scroll to each column in pipeline

dt2patel:
### Related Issues

#762 

Restricted height to just the viewable canvas so that both columns can scroll independently.

Current implementation breaks infinite scrolling. Need to relocate refresh observer.


dt2patel:
closing PR, handled in #771

PR #765, opened 2025-01-23
Improved: Added CSS to make the scrolling of the job configuration section independent of the job list(#762)

R-Sourabh:
### Related Issues

#762 

- Styles the <aside> element to occupy 80% of the viewport height.
- Enables vertical scrolling for content that exceeds the height.
- Uses 80vh to ensure sufficient space for the job list section, allowing for independent scrolling without overlap.
- Adds bottom padding for improved spacing.


src/views/Pipeline.vue
```diff
@@ -821,5 +821,14 @@ ion-chip {
   ion-toolbar > div > ion-chip:first-child {
     margin-left: var(--spacer-3xl);
   }
+  /* Added CSS styles to the <aside> element to occupy 80% of the viewport height, enable vertical scrolling for overflowing content,
+  hide the scrollbar, and add bottom padding for spacing */
+  aside {
+    height: 80vh;
```
dt2patel, line 827: we can't use a fixed number here. Need to take approach similar to my MR where I used 100% height of parent container.

dt2patel (changes requested):

PR #771, opened 2025-02-04
Implemented: Independent scrolling behavior for the job list section and the job config aside elements(#762)

R-Sourabh:
### Related Issues

#762 

- The main element's top margin is removed, its height is set to 100%, and vertical overflow is hidden. Because of this, the main element now takes up the full height of the page.  
- The section and aside elements are allowed to scroll independently, their scrollbars are hidden, and padding is added to their top.  
- Since scrolling is now moved to the section, a scroll event and the `enableScrolling` function have been added to it. These help toggle the `isScrollingEnabled` flag to control whether infinite scrolling can be triggered.  
- Additionally, the class `ion-content-scroll-host` has been added to the section to let `ion-infinite-scroll` recognize the scrolling element.  
- Also updated the space variable for the job cards' margin-block.

### DIFF
--- a/src/views/Pipeline.vue
+++ b/src/views/Pipeline.vue
@@ -30,7 +30,7 @@
       </div>
     </ion-header>
 
-    <ion-content ref="contentRef" :scroll-events="true" @ionScroll="enableScrolling()" id="filter-content">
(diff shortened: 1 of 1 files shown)


dt2patel (approved):
Looks good to me

Issue #840, opened 2025-07-08
Remove Upload refunds job from order section


swati1717:
As we are not using the upload refund job, we need to remove the job from the  order section of the job manager

PR #841, opened 2025-07-08
Removed: UPLD_REFUNDS job from Job Manager App (#840) 

swati1717:
As we are not using the upload refund job, so removed the job from the order section of the job manager

related issue : https://github.com/hotwax/job-manager/issues/840


src/views/Orders.vue
```diff
@@ -110,14 +110,6 @@
               <ion-label v-if="!isLoading" slot="end">{{ getJobScheduleStatus("UPLD_CNCLD_ORDRS", isMaargJobAvailable(jobEnums['UPLD_CNCLD_ORDRS'])) }}</ion-label>
               <ion-skeleton-text v-else style="width: 30%;" animated />
             </ion-item>
-            <ion-item @click="viewJobConfiguration({ id: 'UPLD_REFUNDS', status: getJobStatus(jobEnums['UPLD_REFUNDS'])})" detail button>
-              <ion-label class="ion-text-wrap">
-                {{ translate("Refunds") }}
```
ymaheshwari1, line 115: Check if `Refunds` string is used anywhere in the code with translate and if not remove its entry from en.json file.

ymaheshwari1 (changes requested):

PR #878, opened 2026-03-24
874 aditya

dt2patel:
### Related Issues

#

### DIFF
--- a/src/store/jobs.spec.ts
+++ b/src/store/jobs.spec.ts
@@ -1,0 +1,116 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createPinia, setActivePinia } from "pinia";
+
+const { apiMock } = vi.hoisted(() => ({
+  apiMock: vi.fn()
+}));
+
+vi.mock("@common", () => ({
+  api: apiMock,
+  translate: (value: string) => value
+}));
+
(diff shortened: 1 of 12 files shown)


src/store/jobs.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      expect.objectContaining({
+        jobName: "generate_CreateOrderFeed",
+        serviceName: "co.hotwax.netsuite.OrderServices.generate#CreateOrderFeed",
+        serviceInParameters: [],
+        serviceJobParameters: []
+      })
+    );
+  });
+
+  it("falls back to the catalog job when the detail API has no jobDetail payload", async () => {
```
dt2patel, line 46: I think this part should be removed. we shouldn't actually be calling catalog api on detail page. I was just having an issue where the single job api was not returning details for job that was on the list page

src/views/catalog.vue
```diff
@@ -145,18 +208,51 @@ const filteredJobs = computed(() => {
     filtered = filtered.filter((job: any) => jobIds.includes(job.instanceOfProductId));
   }
 
+  if (selectedStatus.value !== 'ALL') {
```
dt2patel, line 211: we could probably do a better job here and use a real search JS library instead of writing this if else code

PR #895, opened 2026-05-09
Support data document

Banibrata-Manna:


src/store/useDataDocumentGraphStore.ts
ymaheshwari1, line 1: I think we should not name store files with `use` prefix and `Store` suffix, as its recommended to name the store files same as store IDs.

ymaheshwari1 (changes requested):

PR #907, opened 2026-06-01
[codex] Fix Moqui permission loading during login

### LINKED ISSUE #906: Fix Moqui permission loading during login
## Summary
Job Manager login succeeds against local Moqui, but the post-login permission load still calls the legacy root getPermissions endpoint. The active Moqui admin REST API exposes permissions at admin/user/permissions, so login falls back to an error state after credentials are accepted.

## Expected outcome
When VITE_OMS_TYPE is Moqui, Job Manager should load permissions from the Moqui admin user permissions endpoint while preserving the legacy endpoint for non-Moqui deployments.

dt2patel:
## Business summary
Job Manager users on Moqui-backed environments could enter valid credentials but still land in a login failure state because post-login setup used backend/env assumptions that do not hold for the local Moqui dev flow. This PR keeps the login flow aligned with the active backend contract and verifies the rendered credential login path reaches the catalog screen.

Closes #906

## What changed
- Route Moqui permission loads to `GET admin/user/permissions` with pagination query params.
- Preserve the legacy `POST getPermissions` call for non-Moqui deployments.
- Fix the catalog route import casing so the Vite dev app can boot.
- Treat missing `VITE_MAARG_COMPATIBLE_VERSION` as an unconfigured compatibility check instead of forcing legacy redirect.
- Fix semantic version comparison and guard missing `VITE_LEGACY_APP_URL` from throwing during post-login.
- Add focused tests for permission endpoint selection and compatibility behavior.

## Root cause
The credentials step was already using Moqui-aware auth, but `fetchPermissions()` still used the legacy root `getPermissions` endpoint. Runtime validation also showed the dev server was blocked by a route import case mismatch and local `.env.local` omitted compatibility/legacy redirect variables, causing post-login to throw after auth succeeded.

## Validation
- `pnpm --filter job-manager test:unit src/utils/index.spec.ts src/store/user.spec.ts --run`
- `pnpm --filter job-manager build`
- Rendered login smoke test on local Moqui: `http://127.0.0.1:8115/login` -> select local OMS -> submit local dev credentials -> landed on `/catalog` with Job catalog visible.
- Existing saved-session smoke test on `http://[::1]:8105/login` also lands on `/catalog`.

## Notes
- `pnpm --filter job-manager test:unit --run` was not used as the passing gate because existing `jobs.spec.ts` / `systemMessage.spec.ts` baseline issues are unrelated to this change.

### DIFF
--- a/src/router/index.ts
+++ b/src/router/index.ts
@@ -1,6 +1,6 @@
 import { createRouter, createWebHistory } from '@ionic/vue-router';
 import { RouteRecordRaw } from 'vue-router';
-import Catalog from '@/views/Catalog.vue'
+import Catalog from '@/views/catalog.vue'
 import Settings from "@/views/Settings.vue"
 import FileHistory from "@/views/FileHistory.vue"
 import FileDetail from "@/views/FileDetail.vue"
--- a/src/store/user.spec.ts
+++ b/src/store/user.spec.ts
@@ -1,0 +1,100 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createPinia, setActivePinia } from "pinia";
+
+const { apiMock, isMoquiMock, showToastMock } = vi.hoisted(() => ({
+  apiMock: vi.fn(),
+  isMoquiMock: vi.fn(),
+  showToastMock: vi.fn()
+}));
+
+vi.mock("@common", () => ({
+  api: apiMock,
+  commonUtil: {
+    getOmsURL: () => "http://localhost:8080/rest/s1/",
+    hasError: () => false,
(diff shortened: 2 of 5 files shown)


dt2patel:
Closing this draft because current main has already superseded the core permission-loading fix from #906: permissions are now sourced from admin/user/permissions on the Maarg instance, and the catalog casing fix has also landed separately. This branch is stale against main and would drop newer data-document/dev-login work if merged as-is. Any remaining compatibility/legacy redirect hardening from this branch should be reopened as a small fresh PR if still needed.

Issue #912, opened 2026-06-10
Enhance File History with Search, Sorting, Pagination, Metrics and File Information


rachitaarsh-jpg:
## What is the motivation for adding/enhancing this feature?

Added the feature of searching and sorting within the File History page to improve file searching and user experience. Also implemented proper pagination to efficiently handle and navigate large volumes of file history records.

## What are the acceptance criteria?

- [ ] Users can search file history records using relevant search criteria.
- [ ] Users can sort file history records by status, priority and Configuration.
- [ ] Users can input the desired page they want to go to.
- [ ] Pagination is displayed when the number of records exceeds the page size.
- [ ] Users can navigate between pages using pagination controls.
- [ ] Search, sorting, and pagination work together correctly without data inconsistencies.
- [ ] Existing File History functionality remains unaffected.

## Can you complete this feature request by yourself?

- [x] YES
- [ ] NO

## Additional information

Implemented enhancements to improve usability and navigation of File History records, including search, sorting, pagination, metrics visibility and file information display.

PR #915, opened 2026-06-11
912 fix for filehistory pagination sorting searching

rachitaarsh-jpg:
### Related Issues
#912

Created a new Screen for filehistory details, also added pagination, searching and sorting to file history

Added this features and its working fine


rachitaarsh-jpg:
@dt2patel Sir, @ymaheshwari1 Sir could you please review this PR?

dt2patel:
### Heads-up: this PR breaks log-row navigation from Job Detail (stale `FileDetail` route name)

This PR renames the file-detail route from `name: "FileDetail"` → `name: "FileHistoryDetail"` (`src/router/index.ts:79`) and replaces `FileDetail.vue` with `FileHistoryDetail.vue`. But `src/views/JobDetail.vue` still pushes the **old** route name, so clicking a log row on the Job Detail page (`@click="goToLogDetail(log.logId)"`) now silently fails — vue-router emits a `No match for {name: "FileDetail"}` warning and does not navigate.

**`src/views/JobDetail.vue:727`** — `goToLogDetail`:

```diff
 const goToLogDetail = (logId: string | number) => {
-  router.push({ name: "FileDetail", params: { id: logId } });
+  router.push({ name: "FileHistoryDetail", params: { id: logId } });
 };
```

Could you fold this one-liner into this PR? It keeps the rename and its fix together. The `id` param is unchanged (route path `/file-history/:id`, `props: true`).

Verified against this PR's head: after the change, `git grep '"FileDetail"' src` returns nothing, `git grep -w FileDetail src` returns nothing (no leftover imports/usages), and the pushed name matches the registered route at `src/router/index.ts:79`.

src/views/FileHistory.vue
```diff
@@ -107,91 +231,687 @@
...
-async function openFiltersModal() {
-  const filtersModal = await modalController.create({
-    component: LogFilters,
-    componentProps: { appliedFilters: JSON.parse(JSON.stringify(mdmStore.filters)) }
-  })
+const isScrollable = computed(() => {
+  if (!isServerSideSearch.value) {
+    return (pageIndex.value + 1) * PAGE_SIZE < filteredLogs.value.length;
+  }
+  return mdmStore.islogsScrollable;
```
dt2patel: 🔴 **Next never disables on the last page** (server mode). `isScrollable`'s server branch returns `mdmStore.islogsScrollable` = `logs.length > 0 && logs.length < logsCount`; now that the store holds a single 10-row page while `logsCount` is the full total, that's always true. Clicking Next on the final page advances past `pageCount` to an empty page. Use `pageIndex.value + 1 < pageCount.value` instead (`pageCount` is already correct); you can drop the now-unused `islogsScrollable` getter.

src/views/FileHistory.vue
```diff
@@ -6,99 +6,223 @@
...
+              min="1"
+              :max="pageCount"
+              :value="pageIndex + 1"
+              @keyup="validatePageInput($event)"
+              @change="goToPage($event)"
+              class="page-number-input"
+            />
+            <span>/ {{ pageCount }}</span>
+          </div>
+          <ion-button fill="outline" :disabled="!isScrollable" @click="goToNextPage">
```
dt2patel: 🟠 i18n: ~39 new `translate()` keys aren't in `en.json` (e.g. "Total Files", "Success Rate", "Avg Time", "File Details", "Payload", "Copied to clipboard", "No matching rows", "Expand all", …). The i18n linter will fail — please add them all.

src/views/FileHistory.vue
```diff
@@ -107,91 +231,687 @@
...
+  margin: 0;
+}
+
 .file-name {
+  grid-column: span 4;
+  --padding-start: 0;
+  --inner-padding-end: 0;
+}
+
+.legacy-card {
```
dt2patel: 🟠 Dead CSS: a large block of scoped selectors (`.legacy-*`, `.expanded-details`, `.details-*`, `.score-card`, `.badge*`, `.json-header`, `.raw-json-container`, …) has no matching markup — leftovers from the abandoned inline-expansion design (now replaced by `FileHistoryDetail`). Please delete the unreferenced rules.

src/views/FileHistoryDetail.vue
```diff
@@ -0,0 +1,550 @@
...
+        rawText: raw
+      };
+    } catch {
+      // fall through to CSV / text
+    }
+  }
+
+  try {
+    const rows = (await commonUtil.parseCsv(trimmed as any)) as any[];
+    if (rows && rows.length) {
```
dt2patel, line 364: 🟠 Reuse `commonUtil.copyToClipboard` (the established pattern — see `SystemMessageDetailView.vue:157`) instead of importing `@capacitor/clipboard` and reimplementing copy-with-toast.

src/views/FileHistoryDetail.vue
```diff
@@ -0,0 +1,550 @@
...
+  }
+
+  return {
+    ...createPayload(fileName),
+    contentType: "text",
+    rawText: raw
+  };
+}
+
+const getDuration = (start: number, end: number) => {
```
dt2patel: 🟠 `getDuration` is duplicated verbatim here and in `FileHistory.vue`, and hand-rolls the math. Extract one shared helper in `src/utils`/`@common` implemented with Luxon (`Duration`/`DateTime.fromMillis`).

src/views/FileHistoryDetail.vue
```diff
@@ -0,0 +1,550 @@
...
+  const seconds = Math.floor(diff / 1000);
+  const minutes = Math.floor(seconds / 60);
+  const remainingSeconds = seconds % 60;
+  if (minutes === 0) return `${remainingSeconds}s`;
+  return `${minutes}m ${remainingSeconds}s`;
+};
+
+const copyContent = async () => {
+  try {
+    await Clipboard.write({ string: rawText.value });
```
dt2patel: 🟡 `downloadContent` hand-builds a `Blob` + `<a>` + `URL.createObjectURL`. Prefer `saveAs` (file-saver) or the existing `saveDataFile` util to keep download logic consistent across the app.

src/store/mdmConfig.ts
```diff
@@ -93,21 +95,52 @@ export const useMdmConfigStore = defineStore("mdmConfig", {
...
           this.logsCount = 0
         }
-      } catch(err) {
+      } catch (err) {
         logger.error("Failed to fetch logs", err)
       }
     },
+    async fetchDataManagerLogById(logId: string) {
+      try {
+        let resp = await api({
```
dt2patel: ⚪ `let resp` is never reassigned → use `const` (the sibling action `fetchDataManagerFileContent` already does).

src/views/FileHistoryDetail.vue
```diff
@@ -0,0 +1,550 @@
...
+    showToast(translate("Copied to clipboard"));
+  } catch (error) {
```
dt2patel: ⚪ `catch (error)` binds an unused var — use `catch {` to match the other catch blocks in this file.

dt2patel (changes requested):
Really nice work, Rachit — this is a large, ambitious refactor and a lot of it follows our patterns well: no `api` calls in `.vue` files (all in `mdmConfig` store actions), every new action wrapped in try/catch with `logger.error`, response **data** validated instead of status codes, `Promise.all` for independent calls, and `JsonNode` HTML-escapes before `v-html` (good security instinct). 👏

Requesting changes for two real bugs and a set of styleguide items. Bugs: the **Next button never disables on the last page** in server mode, and the route rename left `JobDetail.vue` pushing the removed `FileDetail` name so log navigation silently fails. Styleguide: ~39 new `translate()` keys aren't in `en.json`; a large block of dead CSS remains; reuse `commonUtil.copyToClipboard` / `saveAs` instead of reimplementing; de-dupe `getDuration` into a Luxon-based shared util. The `SystemMessageDetailView.vue` flicker fix is good but is unrelated scope — fine to keep, just flagging. Details inline.

- `src/views/JobDetail.vue`:727 — 🔴 **Route-rename regression.** You renamed the route `FileDetail → FileHistoryDetail` (and deleted `FileDetail.vue`), but `src/views/JobDetail.vue:727` still does `router.push({ name: "FileDetail", ... })` — the only remaining reference to the old name. Navigating to a log from the Job Detail page now silently fails. Please add `JobDetail.vue` to this PR and update it to `"FileHistoryDetail"`.  _(not anchorable inline: file not in this PR)_

src/views/FileHistory.vue
rachitaarsh-jpg: Updated pagination logic and dropped the unused islogsScrollable getter.

src/views/FileHistory.vue
rachitaarsh-jpg: Removed all orphaned legacy CSS rules

src/views/FileHistoryDetail.vue
rachitaarsh-jpg: Refactored to use the shared commonUtil.copyToClipboard.

src/views/FileHistoryDetail.vue
rachitaarsh-jpg: Extracted into a shared Luxon helper in src/utils

src/views/FileHistoryDetail.vue
rachitaarsh-jpg: Switched to using saveAs from file-saver.

src/store/mdmConfig.ts
rachitaarsh-jpg: Changed let to const.

src/views/FileHistoryDetail.vue
rachitaarsh-jpg: Dropped the unused variable and updated to catch {

rachitaarsh-jpg:
@dt2patel Sir thanks so much for the detailed review and the kind words. Really glad to hear the overall refactor aligns well with our patterns. I've just pushed updates that cover everything you caught. Let me know if anything else catches your eye.

Issue #914, opened 2026-06-11
Enhancement Request: Add Copy Button in Message Detail


RANINA12:
**Description:**
In the **Monitor** tab, under the **Message Detail** section, users currently need to manually select and copy message content.

**Proposed Enhancement:**
Add a **Copy** button within the Message Detail section that copies the message content to the clipboard with a single click.

**Expected Behavior:**

* A Copy button is displayed in the Message Detail section.
* Clicking the button copies the complete message content to the clipboard.
* A confirmation message (e.g., "Copied to clipboard") is displayed upon successful copy.

**Benefit:**
This improves usability and allows users to quickly copy message details for debugging, analysis, and sharing without manually selecting the content.

PR #917, opened 2026-06-11
Add copy button in Message Detail Page

RANINA12:
### Related Issues
Enhancement Request: Add Copy Button in Message Detail #914 

#

### Short Description 

Added a Copy button in the System Message Detail view to allow users to quickly copy message content to the clipboard.

This enhancement improves usability by eliminating the need to manually select and copy text, making it easier to reuse, inspect, and share message content.

### Screenshots of Visual Changes before/after 

Before 

Screenshot from 2026-06-10 10-26-29

After

Screenshot from 2026-06-11 14-30-42

### DIFF
--- a/src/views/SystemMessageDetailView.vue
+++ b/src/views/SystemMessageDetailView.vue
@@ -246,29 +246,40 @@
                 </ion-item>
               </template>
 
-              <ion-item-divider color="light">
-                <ion-label>{{ translate("Message Content") }}</ion-label>
-                <div slot="end" class="content-actions">
-                  <ion-button
-                    v-if="downloadableFilePath && !isEditing"
-                    fill="clear"
-                    size="small"
-                    @click="downloadLinkedFile"
-                  >
-                    {{ translate("Download File") }}
-                    <ion-icon slot="start" :icon="downloadOutline" />
-                  </ion-button>
-                  <ion-button
-                    v-if="isEditable && !isEditing"
-                    fill="clear"
-                    size="small"
-                    @click="isEditing = true"
-                  >
-                    {{ translate("Edit Content") }}
-                    <ion-icon slot="start" :icon="addCircleOutline" />
-                  </ion-button>
-                </div>
-              </ion-item-divider>
+         <ion-item-divider color="light">
+           <ion-label>{{ translate("Message Content") }}</ion-label>
+           <div slot="end" class="content-actions">
+             <ion-button
+               v-if="downloadableFilePath && !isEditing"
+               fill="clear"
+               size="small"
+               @click="downloadLinkedFile"
+             >
+               {{ translate("Download File") }}
+               <ion-icon slot="start" :icon="downloadOutline" />
+             </ion-button>
+
+             <ion-button
+               v-if="formattedContent && !isEditing"
+               fill="clear"
+               size="small"
+               @click="commonUtil.copyToClipboard(message?.messageText)"
+             >
+               {{ translate("Copy") }}
+               <ion-icon slot="start" :icon="copyOutline" />
+             </ion-button>
+
+             <ion-button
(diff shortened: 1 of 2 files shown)


src/views/SystemMessageDetailView.vue
```diff
@@ -246,29 +246,40 @@
...
-                    fill="clear"
-                    size="small"
-                    @click="isEditing = true"
-                  >
-                    {{ translate("Edit Content") }}
-                    <ion-icon slot="start" :icon="addCircleOutline" />
-                  </ion-button>
-                </div>
-              </ion-item-divider>
+         <ion-item-divider color="light">
```
ymaheshwari1: Do not change existing code only for indentation or formatting.

ymaheshwari1 (changes requested):

src/views/SystemMessageDetailView.vue
RANINA12: I've corrected the indentation and pushed the changes to the same PR. Could you please take another look?

ymaheshwari1: Still having indentation issues

RANINA12: I have fix it . Please have a look

src/views/SystemMessageDetailView.vue
```diff
@@ -258,6 +258,15 @@
                     {{ translate("Download File") }}
                     <ion-icon slot="start" :icon="downloadOutline" />
                   </ion-button>
+                  <ion-button
+                    v-if="formattedContent && !isEditing"
+                    fill="clear"
+                    size="small"
+                    @click="commonUtil.copyToClipboard(message?.messageText)"
+                  >
+                    {{ translate("Copy") }}
```
dt2patel, line 267: 🔴 i18n: `translate("Copy")` — there's no `Copy` key in `src/locales/en.json`. Add it alongside `"Download File"`/`"Edit Content"`.

src/views/SystemMessageDetailView.vue
```diff
@@ -258,6 +258,15 @@
                     {{ translate("Download File") }}
                     <ion-icon slot="start" :icon="downloadOutline" />
                   </ion-button>
+                  <ion-button
+                    v-if="formattedContent && !isEditing"
+                    fill="clear"
+                    size="small"
+                    @click="commonUtil.copyToClipboard(message?.messageText)"
```
dt2patel: 🟠 **Copies the wrong text.** The `<pre>` shows `formattedContent` (pretty-printed JSON via `JSON.stringify(JSON.parse(content), null, 2)`), but this handler copies the raw `message?.messageText`. For JSON messages the user copies something different from what they see. Copy what's displayed: `@click="commonUtil.copyToClipboard(formattedContent)"`.

src/views/SystemMessageDetailView.vue
```diff
@@ -258,6 +258,15 @@
...
+                  >
```
dt2patel, line 266: ⚪ `copyOutline` breaks the alphabetical order of the ionicons import (it should sit with the other `c…` icons). Low priority — the block is only loosely sorted.

dt2patel (changes requested):
Nice, focused PR — a Copy affordance next to Download/Edit is a sensible UX win, and you did several things right: reused `commonUtil.copyToClipboard` (matching the existing call on line 157) rather than reinventing clipboard logic, wrapped the label in `translate()`, used directive shorthand and `slot="start"`, and kept the diff clean (the earlier indentation-churn concern is gone). 👏

Two things before merge: `translate("Copy")` points at an i18n key that doesn't exist, and the button copies the wrong text — the `<pre>` shows pretty-printed `formattedContent` but the handler copies the raw `messageText`. Details inline.

Issue #918, opened 2026-06-11
Enhancement Request: Configurable Record Limit


kumar-kartikey-choudhary:
Description:
The application currently fetches and displays only 25 records at a time in list views. Users do not have the ability to change the number of records displayed per page, which can make navigation and data analysis inefficient when working with large datasets.

Proposed Enhancement:
Provide a configurable record limit option similar to the functionality available in  WebTools. Users should be able to select the number of records displayed from predefined values.

Expected Behavior:
A record limit selector is available on listing pages.
Users can choose from predefined options such as:
10
20
50
100
200
500
The selected value is applied immediately and the list refreshes to display the specified number of records.
The selected limit is retained during pagination and page refreshes (if applicable).

PR #919, opened 2026-06-11
Added configurable Record limit in Monitor and Remote Systems tab

kumar-kartikey-choudhary:
### Related Issues
Enhancement Request: Configurable Record Limit #918
#

Added a configurable record limit selector in the Monitor and Remote Systems tabs.

Users can now choose the number of records displayed per page from predefined options (10, 20, 50, 100, 200, and 500). The selected limit is applied immediately to the listing, reducing excessive pagination and improving data visibility when working with large datasets.

### Screenshots of Visual Changes  after
image

Before:
1. Fixed display limit of 25 records.
2. No option available to change the number of displayed records.

After:

1. Added a record limit selector.
2. Users can select from 10, 20, 50, 100, 200, or 500 records per page.
3. The listing refreshes automatically based on the selected limit.

### DIFF
--- a/src/views/SystemMessageMonitor.vue
+++ b/src/views/SystemMessageMonitor.vue
@@ -100,7 +100,7 @@
         <div class="pagination">
           <div class="limit-container">
             <ion-select
-              :label="translate('Limit')"
+              :label="translate('pageSize')"
               label-placement="start"
               interface="popover"
               :value="pageSize"
@@ -158,7 +158,7 @@ import SystemMessageList from "@/components/SystemMessageList.vue";
 import { useSystemMessageStore } from "@/store/systemMessage";
 import { useUtilStore } from "@/store/util";
 
-const pageSize = ref(25);
+const pageSize = ref(20);
 
 const store = useSystemMessageStore();
 const utilStore = useUtilStore();
--- a/src/views/SystemMessageMonitor.vue
+++ b/src/views/SystemMessageMonitor.vue
@@ -98,13 +98,31 @@
         </ion-card>
 
         <div class="pagination">
-          <ion-button fill="outline" :disabled="pageIndex === 0" @click="goToPreviousPage">
-            {{ translate("Previous") }}
-          </ion-button>
-          <ion-note color="medium">{{ translate("Page") }} {{ pageIndex + 1 }} / {{ pageCount }}</ion-note>
-          <ion-button fill="outline" :disabled="pageIndex >= pageCount - 1" @click="goToNextPage">
-            {{ translate("Next") }}
-          </ion-button>
+          <div class="limit-container">
+            <ion-select
+              :label="translate('Limit')"
+              label-placement="start"
+              interface="popover"
+              :value="pageSize"
+              @ionChange="pageSize = $event.detail.value"
+            >
+              <ion-select-option :value="10">10</ion-select-option>
+              <ion-select-option :value="20">20</ion-select-option>
+              <ion-select-option :value="50">50</ion-select-option>
+              <ion-select-option :value="100">100</ion-select-option>
+              <ion-select-option :value="200">200</ion-select-option>
+              <ion-select-option :value="500">500</ion-select-option>
+            </ion-select>
+          </div>
+          <div class="pagination-buttons">
+            <ion-button fill="outline" :disabled="pageIndex === 0" @click="goToPreviousPage">
+              {{ translate("Previous") }}
+            </ion-button>
+            <ion-note color="medium">{{ translate("Page") }} {{ pageIndex + 1 }} / {{ pageCount }}</ion-note>
+            <ion-button fill="outline" :disabled="pageIndex >= pageCount - 1" @click="goToNextPage">
+              {{ translate("Next") }}
+            </ion-button>
+          </div>
         </div>
         <SystemMessageList
           :messages="messages"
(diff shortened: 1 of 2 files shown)


src/views/SystemMessageMonitor.vue
```diff
@@ -140,7 +158,7 @@ import SystemMessageList from "@/components/SystemMessageList.vue";
 import { useSystemMessageStore } from "@/store/systemMessage";
 import { useUtilStore } from "@/store/util";
 
-const PAGE_SIZE = 25;
+const pageSize = ref(25);
```
dt2patel: 🔴 **Default isn't selectable.** `pageSize = ref(25)`, but `25` isn't one of the `ion-select-option`s (10/20/50/100/200/500), so the Limit control renders **blank** on load while the list actually fetches 25. Pick a default that exists (e.g. `ref(20)` or `ref(50)`), or add a `25` option — and keep it identical in both views.

src/views/SystemMessageRemoteDetail.vue
```diff
@@ -151,7 +169,7 @@ import { caretBackOutline, caretForwardOutline } from "ionicons/icons";
 // Type based declaration
 const props = defineProps<{ id?: string }>();
 
-const PAGE_SIZE = 25;
+const pageSize = ref(25);
```
dt2patel: 🔴 Same blank-selector bug: `ref(25)` isn't in the dropdown. Use a value that's in the options, matching `SystemMessageMonitor.vue`.

src/views/SystemMessageMonitor.vue
```diff
@@ -98,13 +98,31 @@
...
-          <ion-button fill="outline" :disabled="pageIndex === 0" @click="goToPreviousPage">
-            {{ translate("Previous") }}
-          </ion-button>
-          <ion-note color="medium">{{ translate("Page") }} {{ pageIndex + 1 }} / {{ pageCount }}</ion-note>
-          <ion-button fill="outline" :disabled="pageIndex >= pageCount - 1" @click="goToNextPage">
-            {{ translate("Next") }}
-          </ion-button>
+          <div class="limit-container">
+            <ion-select
+              :label="translate('Limit')"
```
dt2patel: 🟠 i18n: `translate('Limit')` — the key `Limit` doesn't exist in `en.json`. We already have a `Page Size` key for this exact concept (used in `DataDocumentQueryPreview.vue`) — please reuse `translate('Page Size')`.

src/views/SystemMessageRemoteDetail.vue
```diff
@@ -92,13 +92,31 @@
...
-                <ion-button fill="clear" slot="icon-only" :disabled="pageIndex === 0" @click="goToPreviousPage">
-                  <ion-icon :icon="caretBackOutline" />
-                </ion-button>
-                <ion-note color="medium">{{ pageIndex + 1 }} / {{ pageCount }}</ion-note>
-                <ion-button fill="clear" slot="icon-only" :disabled="pageIndex >= pageCount - 1" @click="goToNextPage">
-                  <ion-icon :icon="caretForwardOutline" />
-                </ion-button>
+                <div class="limit-container">
+                  <ion-select
+                    :label="translate('Limit')"
```
dt2patel: 🟠 Same missing `Limit` key — reuse `translate('Page Size')`.

src/views/SystemMessageMonitor.vue
```diff
@@ -248,9 +269,21 @@ onIonViewWillEnter(async () => {
...
   padding: 16px;
 }
+
+.pagination-buttons {
+  display: flex;
+  align-items: center;
+  gap: 12px;
+}
+
+.limit-container {
```
dt2patel: 🟡 The Limit `ion-select` block and the `.limit-container`/`.pagination-buttons` scoped CSS are duplicated verbatim across both views and will drift over time. Extract a small shared `PageSizeSelector` component (props: `modelValue` + `options`), or at minimum a shared options constant.

src/views/SystemMessageRemoteDetail.vue
```diff
@@ -260,29 +278,34 @@ const goToPreviousPage = () => {
 const goToNextPage = () => {
   pageIndex.value += 1;
 };
-
-watch([queryString, selectedStatusId], () => {
-  pageIndex.value = 0
-})
-
-watch([pageIndex], async () => {
+const fetchRelatedMessages = async () => {
```
dt2patel, line 245: ⚪ Missing terminating semicolons on the payload object and the `await` call inside `fetchRelatedMessages()` — keep it consistent with the rest of the block.

dt2patel (changes requested):
Nice work, Kartikey — the reactivity is correct (`pageSize` is a primitive so `ref()` is right), it's properly threaded through `pageCount` and the fetch payloads, and the watcher refactor in `SystemMessageRemoteDetail.vue` (extracting `fetchRelatedMessages()` and guarding the page-reset) thoughtfully avoids a double-fetch. 👏

Requesting changes on three things: the default `ref(25)` isn't one of the dropdown options so the Limit selector renders blank on load; the `'Limit'` i18n key doesn't exist (we already have `'Page Size'`); and the selector markup/CSS/watcher is copy-pasted across both views and will drift. One note so you're not over-penalized: the single quotes in `translate('Limit')` are correct — that's the right style for a string nested in a double-quoted attribute.

dt2patel:
holding new feature for now. only merging bug fixes

PR #921, opened 2026-06-11
refactor: moved legacy app link to settings page

harshmahajan7:
Moved the Legacy App access from the main sidebar into the Settings page under the App section. It makes more sense there and cleans up the sidebar navigation.

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -14,20 +14,11 @@
             button
             router-direction="root"
             :router-link="page.url"
-            class="hydrated"
             :class="{ selected: selectedIndex === index }">
             <ion-icon slot="start" :ios="page.iosIcon" :md="page.mdIcon" />
             <ion-label>{{ translate(page.title) }}</ion-label>
           </ion-item>
-          <ion-item
-            v-else-if="page.action"
-            button
-            @click="page.action()"
-            class="hydrated"
-            :class="{ selected: selectedIndex === index }">
-            <ion-icon slot="start" :ios="page.iosIcon" :md="page.mdIcon" />
-            <ion-label>{{ translate(page.title) }}</ion-label>
-          </ion-item>
+
           <ion-item-divider color="light" v-else>
             <ion-label color="medium">{{ translate(page.title) }}</ion-label>
           </ion-item-divider> 
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -102,7 +102,6 @@ import {
   constructOutline,
   documentTextOutline,
   fileTrayStackedOutline,
-  gitNetworkOutline,
   globeOutline,
   pulseOutline,
   searchCircleOutline,
@@ -122,7 +121,6 @@ const userProfile = computed(() => userStore.getUserProfile);
 
 const browserTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
 
-// Filtering array of app pages, retaining only those elements (pages) that have the necessary permissions for display.
 const getValidMenuItems = (appPages: any) => {
   return appPages.filter(
     (appPage: any) =>
@@ -203,13 +201,7 @@ let appPages = [
     mdIcon: documentTextOutline,
     childRoutes: ["/data-documents/"],
   },
-  // {
-  //   title: "Feeds",
-  //   url: "/data-document-feeds",
-  //   iosIcon: gitNetworkOutline,
-  //   mdIcon: gitNetworkOutline,
-  //   childRoutes: ["/data-document-feeds/"]
-  // },
+
   {
     title: "Export History",
     url: "/data-document-export-history",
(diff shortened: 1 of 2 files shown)


src/components/Menu.vue
```diff
@@ -26,6 +19,15 @@
             <ion-icon slot="start" :ios="page.iosIcon" :md="page.mdIcon" />
             <ion-label>{{ translate(page.title) }}</ion-label>
           </ion-item>
+          <ion-item
```
dt2patel: 🟠 **Dead branch.** This `v-else-if="page.action"` block can never render — no entry in `appPages` defines an `action` key (search the file for `action:` → no matches). Since this PR moves the legacy action *out* of the menu, please delete the whole block; the loop only needs the `v-if="page.url"` item and the `v-else` divider.

src/components/Menu.vue
```diff
@@ -26,6 +19,15 @@
...
+            v-else-if="page.action"
+            button
+            @click="page.action()"
+            class="hydrated"
```
dt2patel: 🟡 `class="hydrated"` is a class Ionic/Stencil applies at runtime — don't author it by hand. (Moot if you remove this branch per the note above.)

src/views/Settings.vue
```diff
@@ -244,6 +259,7 @@ import Image from '@/components/Image.vue'
 import { cookieHelper, commonUtil, translate } from '@common';
 import { useAuth } from '@common/composables/useAuth';
 import { DateTime } from 'luxon';
+import { redirectToLegacyApp } from '@/utils';
```
dt2patel: ⚪ Team standard is double quotes for imports (`from "@/utils"`). The rest of this file already uses single quotes, so matching it is acceptable — flagging for awareness on new code only.

dt2patel (changes requested):
Nice work on the core goal — the legacy-app link is cleanly relocated from the Menu sidebar into a dedicated Settings card, the redirect logic is correctly reused from `@/utils` (good, no reimplementation), all visible text is wrapped in `translate()`, and the diff is tight and focused. 👏

Requesting changes for a few cleanup items: two dead imports left behind in `Menu.vue`, a `v-else-if="page.action"` branch that can never render (no `appPages` entry defines `action`), and two new i18n keys that aren't in `en.json`. Details are inline. None are large — fix these and it's good to go.

- `src/components/Menu.vue`:71 — 🟠 **Unused import.** `redirectToLegacyApp`'s only call site (the old legacy-app `ion-item`) was removed in this PR, so this import is now dead. Please delete it — styleguide: dead code must be removed before merge.  _(not anchorable inline: line 71 not in diff hunk)_
- `src/components/Menu.vue`:66 — 🟠 **Unused import.** `openOutline` was only used by the removed legacy `ion-item` icon and is now unreferenced. Remove it from this destructured import.  _(not anchorable inline: line 66 not in diff hunk)_
- `src/locales/en.json` — 🟠 i18n: `translate("Go to Legacy App")` and the `"Access the legacy application…"` description aren't in `src/locales/en.json`. Every `translate()` key must exist in the locale file — please add both. (`"Legacy App"` already exists, so that one's fine.)  _(not anchorable inline: file not in this PR)_

harshmahajan7:
Hi @dt2patel, I've pushed the requested changes:
- Removed the dead `v-else-if="page.action"` block from `Menu.vue`
- Removed the manual `class="hydrated"` assignment from `Menu.vue`
- Updated the import statement in `Settings.vue` to use double quotes (`import { redirectToLegacyApp } from "@/utils";`) per the style guide

Please let me know if anything else is needed!

src/components/Menu.vue
ymaheshwari1: Unused import left as its not used anymore.

ymaheshwari1 (changes requested):

ymaheshwari1:
@dt2patel Sir, this feature will not work, as on JUNE-15 branch we have only MOQUI login and thus we won't have oms, which will cause moving to legacy app fails.

ymaheshwari1:
Converting this to draft for now, so that it does not gets merged by mistake, until the above point is not resolved.

src/components/Menu.vue
```diff
@@ -39,13 +42,28 @@
...
         <ion-item v-if="userProfile.stores?.length > 2" lines="none">
           <!-- WHY EVENTS ($emit) IS USED WITH ION CHANGE: https://michaelnthiessen.com/pass-function-as-prop/ -->
-          <ion-select interface="popover" :value="currentProductStore.productStoreId" @ionChange="setProductStore($event.target.value)">
-            <ion-select-option v-for="store in (userProfile ? userProfile.stores : [])" :key="store.productStoreId" :value="store.productStoreId" >{{ store.storeName || store.productStoreId }}</ion-select-option>
+          <ion-select
+            interface="popover"
+            :value="currentProductStore.productStoreId"
+            @ionChange="setProductStore($event.target.value)"
+          >
+            <ion-select-option
```
ymaheshwari1, line 61: Do not make formatting specific changes in this PR.

ymaheshwari1 (changes requested):

src/components/Menu.vue
harshmahajan7: Ok sir

PR #923, opened 2026-06-12
Fix: Apply reverse pagination for system messages

### LINKED ISSUE #922: System messages are sorted oldest to newest instead of newest to oldest
**Describe the bug**
On the System Messages page (`/system-messages`), the messages are fetched and displayed in sequential order (oldest to newest) because the remote API backend does not inherently support the `orderByField` parameter. Users expect to see the most recent system messages at the top of the first page.

**To Reproduce**
1. Navigate to `/system-messages` in the application.
2. Note the timestamps of the messages on the first page.
3. Observe that they are the oldest messages in the database rather than the newest.

**Expected behavior**
The system messages should be sorted globally by time, from newest to oldest, ensuring the very first page contains the most recently generated system messages.

**Proposed Solution**
Implement a "Reverse Pagination" algorithm in the frontend Vue store (`systemMessage.ts`) that:
1. Performs a pre-flight count request to determine the total number of pages.
2. Mathematically calculates the reverse backend page index (fetching the last backend page when the frontend requests the first page).
3. Reverses the returned array locally.
4. Includes a safe fallback to standard pagination if the pre-flight request fails.

harshmahajan7:
Resolves #922

### Description
This pull request introduces a reverse-pagination mechanism in the frontend to correctly display system messages from newest to oldest. Since the remote API backend (`dev-oms`) ignores the `orderByField` parameter, this implementation:
1. Performs a pre-flight count request to calculate total pages based on current filters.
2. Inverts the `pageIndex` internally to request the trailing elements.
3. Locally reverses the items upon retrieval.
4. Includes a safe `try/catch` fallback to original sequence if the pre-flight fails.

### Testing
- Tested against remote instances that ignore URL ordering.
- Checked edge cases with filters applied.
- Verified smooth fallback functionality.

### DIFF
--- a/src/store/systemMessage.ts
+++ b/src/store/systemMessage.ts
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
       this.loading = true;
 
       try {
+        const pageSize = Number(payload.pageSize ?? 25);
+        let frontendPageIndex = Number(payload.pageIndex ?? 0);
+        let fetchPageIndex = frontendPageIndex;
+        let isReversed = false;
+
+        // Safely attempt reverse pagination hack
+        try {
+          const countPayload = { ...payload };
+          delete countPayload.pageIndex;
+          delete countPayload.pageSize;
+          
+          const countResponse = await api({
+            url: "admin/systemMessages",
+            method: "GET",
+            params: { pageSize: 1, pageIndex: 0, ...countPayload }
+          });
+          
+          const totalCount = countResponse.data?.systemMessagesCount;
+          if (totalCount > 0) {
+            const totalPages = Math.ceil(totalCount / pageSize);
+            fetchPageIndex = totalPages - 1 - frontendPageIndex;
+            if (fetchPageIndex < 0) fetchPageIndex = 0;
+            isReversed = true;
+          }
+        } catch (countErr) {
+          logger.warn("Failed to fetch system messages count for reverse pagination, falling back to normal pagination.", countErr);
+        }
+
         const response = await api({
           url: "admin/systemMessages",
           method: "GET",
           params: {
-            pageSize: Number(payload.pageSize ?? 25),
-            pageIndex: Number(payload.pageIndex ?? 0),
-            orderBy: "-initDate",
-            ...payload
+            ...payload,
+            pageSize,
+            pageIndex: fetchPageIndex,
+            orderByField: "-initDate"
           }
         });
 
         if(response.data?.systemMessagesCount) {
-          this.systemMessages = response.data.systemMessages;
+          if (isReversed) {
+            this.systemMessages = response.data.systemMessages.reverse();
+          } else {
+            this.systemMessages = response.data.systemMessages;
+          }
           this.systemMessageTotal = response.data.systemMessagesCount;
           return;
         }
--- a/src/store/systemMessage.ts
+++ b/src/store/systemMessage.ts
@@ -119,7 +119,7 @@ export const useSystemMessageStore = defineStore("systemMessage", {
         const pageSize = Number(payload.pageSize ?? 25);
         const pageIndex = Number(payload.pageIndex ?? 0);
 
-        // Strip pagination params for the API call
+        // Strip pagination params for API calls
         const requestPayload = { ...payload };
         delete requestPayload.pageIndex;
         delete requestPayload.pageSize;
@@ -146,7 +146,8 @@ export const useSystemMessageStore = defineStore("systemMessage", {
           const endIndex = startIndex + pageSize;
           
           this.systemMessages = sortedMessages.slice(startIndex, endIndex);
-          this.systemMessageTotal = response.data.systemMessagesCount ?? sortedMessages.length;
+          // Set total to the length of our local sorted array so pagination doesn't break
+          this.systemMessageTotal = sortedMessages.length;
           return;
         }
 
--- a/src/store/systemMessage.ts
+++ b/src/store/systemMessage.ts
@@ -32,6 +32,7 @@ export const useSystemMessageStore = defineStore("systemMessage", {
     relatedMessages: [] as any[],
     linkedMessages: [] as any[],
     systemMessageTotal: 0,
+    lastFilters: "",
     loading: false,
     enums: [] as any[],
     currentSystemMessageStatusHistory: []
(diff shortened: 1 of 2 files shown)


harshmahajan7:
@dt2patel Sir, @ymaheshwari1 Sir could you please review this PR?

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
+          const countResponse = await api({
+            url: "admin/systemMessages",
+            method: "GET",
+            params: { pageSize: 1, pageIndex: 0, ...countPayload }
+          });
+          
+          const totalCount = countResponse.data?.systemMessagesCount;
+          if (totalCount > 0) {
+            const totalPages = Math.ceil(totalCount / pageSize);
+            fetchPageIndex = totalPages - 1 - frontendPageIndex;
```
dt2patel: 🔴 **Page math breaks on a partial last page.** `fetchPageIndex = totalPages - 1 - frontendPageIndex` maps frontend page 0 onto the *last* server page, which is the partial one. 30 items @ pageSize 25 → `totalPages = 2` → frontend page 0 shows server page 1 = only **5 rows** (should be 25); frontend page 1 shows 25. The per-page counts invert. Abandon the index-inversion and sort client-side instead.

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
+            pageSize,
+            pageIndex: fetchPageIndex,
+            orderByField: "-initDate"
           }
         });
 
         if(response.data?.systemMessagesCount) {
-          this.systemMessages = response.data.systemMessages;
+          if (isReversed) {
+            this.systemMessages = response.data.systemMessages.reverse();
```
dt2patel: 🔴 `.reverse()` only reorders rows *within* the single returned page, so global newest-first order across page boundaries is still wrong — global item 25 should follow item 24 but lands on a different frontend page entirely. Sort the **full** result set by `initDate` (desc) before slicing into pages.

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
           method: "GET",
           params: {
-            pageSize: Number(payload.pageSize ?? 25),
-            pageIndex: Number(payload.pageIndex ?? 0),
-            orderBy: "-initDate",
-            ...payload
+            ...payload,
+            pageSize,
+            pageIndex: fetchPageIndex,
+            orderByField: "-initDate"
```
dt2patel: 🟠 `orderBy → orderByField`: this endpoint ignores **both** params server-side (that's the real reason ordering never worked and why this hack exists), so the rename is a no-op — and it now disagrees with the sibling call in `dataDocuments.ts` which uses `orderBy: "-initDate"`. Drop the reliance on a server order param entirely.

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
+        let fetchPageIndex = frontendPageIndex;
+        let isReversed = false;
+
+        // Safely attempt reverse pagination hack
+        try {
+          const countPayload = { ...payload };
+          delete countPayload.pageIndex;
+          delete countPayload.pageSize;
+          
+          const countResponse = await api({
```
dt2patel: 🟠 The pre-flight count request (`pageSize: 1`) fires on **every** `fetchSystemMessages` call just to compute `totalPages`, doubling network round-trips — the main response already returns `systemMessagesCount`. Remove it.

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
       this.loading = true;
 
       try {
+        const pageSize = Number(payload.pageSize ?? 25);
+        let frontendPageIndex = Number(payload.pageIndex ?? 0);
```
dt2patel: ⚪ `frontendPageIndex` is never reassigned → use `const`.

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
+        let fetchPageIndex = frontendPageIndex;
+        let isReversed = false;
+
+        // Safely attempt reverse pagination hack
+        try {
+          const countPayload = { ...payload };
+          delete countPayload.pageIndex;
+          delete countPayload.pageSize;
+
```
dt2patel: ⚪ Trailing whitespace on this blank line.

src/store/systemMessage.ts
```diff
@@ -115,19 +115,51 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
+          const countPayload = { ...payload };
+          delete countPayload.pageIndex;
+          delete countPayload.pageSize;
+          
+          const countResponse = await api({
+            url: "admin/systemMessages",
+            method: "GET",
+            params: { pageSize: 1, pageIndex: 0, ...countPayload }
+          });
+
```
dt2patel: ⚪ Trailing whitespace on this blank line.

dt2patel (changes requested):
Thanks for tackling this, Harsh — the intent (newest messages first) is right, and there are good habits here: the count fetch has its own try/catch with a graceful fallback and `logger.warn` (not `console.log`), the outer fetch keeps its try/catch + state reset, and you pulled `pageSize` into a single computed.

I'm requesting changes because the core approach has correctness bugs and diverges from the pattern we already use for this exact endpoint — let's **rework it rather than patch it**. The reverse-index scheme misaligns rows whenever the total isn't a multiple of `pageSize`, and `.reverse()` only reorders *within* a page so cross-page order stays wrong. This endpoint (`GET admin/systemMessages`) ignores `orderBy`/`orderByField` server-side anyway. The intended pattern is the sibling `fetchExportHistory` in `dataDocuments.ts`: fetch with a generous `pageSize` and **sort client-side by `initDate` descending** (using our Luxon/`@common` date helpers), then paginate the sorted array — no count round-trip, no index math. Details inline.

- `src/store/systemMessage.ts`:170 — ⚪ Missing semicolon: `this.systemMessages = [];`  _(not anchorable inline: line 170 not in diff hunk)_

harshmahajan7:
I have verified these changes locally and the client-side sorting and pagination are working perfectly now. Thanks again for the review!

src/store/systemMessage.ts
```diff
@@ -115,27 +117,94 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
+        if (totalCount === 0) {
+          this.systemMessages = [];
           return;
         }
 
-        throw new Error(response.data);
+        // 2. Calculate which backend items we need
+        // Frontend wants: newest first. Item 0 is the absolute newest in the DB.
+        // Backend provides: oldest first. Item 0 is the absolute oldest.
+        let frontendStart = pageIndex * pageSize;
```
ymaheshwari1: We do not need all this logic in the app, instead we can simply add support of orderBy field in backend in search-form-inputs, so it automatically handles the sorting of messages.

ymaheshwari1 (changes requested):

src/store/systemMessage.ts
```diff
@@ -119,20 +119,19 @@ export const useSystemMessageStore = defineStore("systemMessage", {
...
         });
 
-        if(response.data?.systemMessagesCount) {
+        if (response.status === 200 && response.data?.systemMessages) {
           this.systemMessages = response.data.systemMessages;
-          this.systemMessageTotal = response.data.systemMessagesCount;
-          return;
+          this.systemMessageTotal = response.data.systemMessagesCount || 0;
+        } else {
+          throw new Error(response.data);
```
ymaheshwari1, line 133: I think we do not need all this changes.

ymaheshwari1 (changes requested):

harshmahajan7:
No changes are required in the frontend (`job-manager`) because the store is already sending `orderBy: "-initDate"` in the API request parameters. 

By adding support for the `orderBy` parameter directly in the backend (`maarg-util` -> `UtilityServices.xml`), the sorting (newest-to-oldest) is now handled entirely on the backend database level without needing any frontend modifications.

Issue #945, opened 2026-06-16
Bug: File History and System Message screens fail to load due to route injection error


rachitaarsh-jpg:
## Current behavior
File History and System Message Monitor pages crash on load. KPIs show `0` and data fails to fetch. The console throws:
`[Vue warn]: injection "Symbol(route location)" not found.`
`TypeError: Cannot read properties of undefined (reading 'query')`

## Expected behavior
Pages should load successfully, parse URL queries, and fetch data without crashing.

## Steps to reproduce the issue
1. Start the dev server (`npm run dev`).
2. Navigate to `File history` or `System Message Monitor`.
3. Observe empty data and console errors.

## Can you handle fixing this bug by yourself?
- [x] YES
- [ ] NO

## Environment details
- Browser: Chrome 
- OS: Linux (Ubuntu)
- Code Version: `main` branch 

## Additional information
The crash is caused by `useRoute()` returning `undefined` during component setup in Ionic. The fix is to remove `useRoute()` and read directly from `router.currentRoute.value.query` in `onIonViewWillEnter`.

dt2patel:
Code triage note (high confidence): likely root cause is use of `useRoute()` inside Ionic cached views where Vue Router injection is not always available during setup.

Evidence:
- `FileHistory.vue` and `SystemMessageMonitor.vue` read `route.query` from `useRoute()`.
- The reported crash is `injection "Symbol(route location)" not found` followed by `Cannot read properties of undefined (reading 'query')`.
- PR https://github.com/hotwax/job-manager/pull/946 appears to implement the right direction by reading `router.currentRoute.value.query` inside Ionic lifecycle instead.

Proposed fix:
- Remove `useRoute()` from the affected Ionic views.
- Read query params from `router.currentRoute.value.query` inside `onIonViewWillEnter` / refresh paths.
- Apply the same pattern to any other Ionic views found using `useRoute()` during setup.

Validation:
1. Load File History directly and via sidebar navigation with and without query params.
2. Load System Message Monitor directly and via sidebar navigation with and without query params.
3. Confirm KPIs fetch and list data loads.
4. Confirm the console has no `Symbol(route location)` warning or `route.query` crash.

rachitaarsh-jpg:
[...]

PR #946, opened 2026-06-16
Fix Vue Router injection error 

rachitaarsh-jpg:
### Related Issues
#945 

This PR resolves app crashes and build errors by fixing the following:
- **Vue Router Crash:** Replaces `useRoute()` with `router.currentRoute.value` in components (`FileHistory`, `Catalog`, `SystemMessageMonitor`, etc.). This fixes injection context errors (`Symbol(route location) not found`) caused by calling `useRoute`.

### DIFF
--- a/src/views/Catalog.vue
+++ b/src/views/Catalog.vue
@@ -157,6 +157,7 @@ import CreateJobModal from '@/components/CreateJobModal.vue';
 import { useJobStore } from '@/store/jobs';
 
 const jobStore = useJobStore();
+const route = router.currentRoute.value;
 
 const jobs = computed(() => jobStore.getJobs)
 const categories = computed(() => jobStore.getCategories)
@@ -165,7 +166,6 @@ const categoryRollups = computed(() => jobStore.getCategoryRollups)
 
 onIonViewWillEnter(async () => {
   emitter.on("productStoreUpdated", jobStore.fetchJobs)
-  const route = router.currentRoute.value;
   if (route.query?.status) {
     selectedStatus.value = route.query.status as string;
   } else {
--- a/src/views/Catalog.vue
+++ b/src/views/Catalog.vue
@@ -165,8 +165,9 @@ const categoryRollups = computed(() => jobStore.getCategoryRollups)
 
 onIonViewWillEnter(async () => {
   emitter.on("productStoreUpdated", jobStore.fetchJobs)
-  if (router.currentRoute.value.query?.status) {
-    selectedStatus.value = router.currentRoute.value.query.status as string;
+  const route = router.currentRoute.value;
+  if (route.query?.status) {
+    selectedStatus.value = route.query.status as string;
   } else {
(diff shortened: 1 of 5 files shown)


src/views/DataDocumentFeedDetail.vue
```diff
@@ -150,12 +150,12 @@ import {
...
 import { computed, ref } from "vue";
-import { useRoute, useRouter } from "vue-router";
+import { useRouter } from "vue-router";
 
 import { translate } from "@common";
 import { useDataDocumentStore } from "@/store/dataDocuments";
 
-const route = useRoute();
+
 const router = useRouter();
```
Banibrata-Manna: Please update this as well

rachitaarsh-jpg: Updated

dt2patel (approved):
Approved. I reviewed the final head (78bdac4d): the change is narrowly scoped to removing useRoute() from the affected Ionic views and reading route state through the app router inside the lifecycle paths. Verified the PR head builds successfully in an isolated AccxUI-style workspace; GitHub build checks are also green. I do not see a blocking issue in the remaining advisory comments.

src/views/Catalog.vue
```diff
@@ -167,8 +165,8 @@ const categoryRollups = computed(() => jobStore.getCategoryRollups)
 
 onIonViewWillEnter(async () => {
   emitter.on("productStoreUpdated", jobStore.fetchJobs)
-  if (route.query.status) {
-    selectedStatus.value = route.query.status as string;
+  if (router.currentRoute.value.query?.status) {
```
ymaheshwari1: Instead of using router.currentRoute.value on multiple places, update line 160 to define it once like

const route = router.currentRoute.value

And after this no other change will be required in the file.

ymaheshwari1 (changes requested):

src/views/Catalog.vue
rachitaarsh-jpg: Sir i have fixed this issue in this files and other files as well. Kindly provide a review for this.

src/views/Catalog.vue
```diff
@@ -167,7 +165,8 @@ const categoryRollups = computed(() => jobStore.getCategoryRollups)
 
 onIonViewWillEnter(async () => {
   emitter.on("productStoreUpdated", jobStore.fetchJobs)
-  if (route.query.status) {
+  const route = router.currentRoute.value;
```
ymaheshwari1: This needs to be done outside of onIonViewWillEnter on the same line as it was previously defined

ymaheshwari1 (changes requested):

src/views/Catalog.vue
rachitaarsh-jpg: Sir, I have changed this as you asked, kindly check it.

ymaheshwari1 (approved):

PR #951, opened 2026-06-17
Fix: Prevent premature field creation in Data Document Builder (#950)

### LINKED ISSUE #950: Data Document Builder: Selected field is not persisted for newly added Field or Related Entity
## Current behavior

When adding a **Field** or a **Related Entity** in the Data Document Builder, selecting a field and clicking **Save All** does not correctly persist the selection.

* For **Add Field**: Clicking **Add Field** immediately creates a blank placeholder entry in the graph before any field is selected. If the modal is closed without selecting a field, the blank entry remains. The selected field may not be persisted correctly after **Save All**.

* For **Add Related Entity**: Clicking **Add Related Entity**, selecting a related entity, and selecting one of its fields results in a new related entity group being created, but the selected field is not associated with the group. The entry appears blank after **Save All**.

  In the graph visualization, only the field count is incremented when a related entity field is selected. The relationship between the parent entity and the related entity is not reflected in the graph, and the corresponding entity block in the bottom section is created as an empty entry. As a result, the graph representation becomes inconsistent with the persisted Data Document structure.

---

## Expected behavior

* Clicking **Add Field** should **not** create any entry. An entry should only be created after a field is selected.

* Clicking **Add Related Entity** should **not** create any entry. An entry containing the selected field should only be created after selecting both a related entity and a field.

* Cancelling either modal without making a selection should not create any entry.

* Related entity relationships should be reflected correctly in the graph visualization.

* The corresponding entity block in the bottom section should display the selected related entity field instead of appearing as an empty entry.

* Field counts displayed in graph nodes should remain synchronized with the fields configured in the Data Document.

---

2. Create a new Data Document with the following details:

   * **Data Document ID:** `CompletedReturnItems`
   * **Primary Entity:** `org.apache.ofbiz.order.return.ReturnItem`
   * **Name:** `Completed Return Items`
   * **Title:** `${returnId}:${returnItemSeqId}`

3. Click **Save All** to create the Data Document.

4. In the bottom section of the Data Document Builder, there are two actions available:

   * **Add Field**
   * **Add Related Entity**

5. Click **Add Field** — observe that a blank entry appears immediately, before selecting anything.

6. Select a field. Click **Save All** — observe whether the field is correctly persisted.

7. Click **Add Related Entity**.

8. Select any related entity.

9. Select any field from the related entity.

10. Click **Save All** — observe that the related entity group appears as an empty entry with no field inside.

---

### Empty related entity created after saving

Image

---

* [x] YES
* [ ] NO

---

## Environment details

* Browser: Chrome 137.x
* OS: Ubuntu 24.04 LTS
* Code Version: `main` (fixed in `fix/data-document-related-entity-blank-entry` branch)

---

## Additional information

Both the **Add Field** and **Add Related Entity** flows share the same root cause.

The original implementation in `openNewFieldModal` eagerly called `graphStore.addFieldPath()` before the user selected anything, creating a blank placeholder in the graph. The index of this placeholder was then resolved using `item.fieldSeqId === field?.fieldSeqId`, but this comparison always failed because `projectDataDocumentGraph` assigns synthetic sequence IDs (for example `"10"` and `"20"`) to new fields while the raw record holds `fieldSeqId: ""`.

As a result, `activeFieldIndex` was always `-1`, and `selectField` never updated the placeholder, leaving blank entries in both the graph and the bottom entity section.

The fix defers field creation entirely. No entry is added until the user explicitly selects a field, at which point the correct full path is constructed and `addFieldPath()` is called exactly once with the resolved field path.

This fix also resolves graph visualization inconsistencies.

Previously:

* Adding a Related Entity increased the field count in the graph.
* Relationships between parent and related entities were not displayed in the graph.
* The corresponding entity block in the bottom section was created as an empty entry.
* Graph nodes could become inconsistent with the persisted Data Document structure.

Now:

* Related entity relationships are reflected correctly in the graph.
* The corresponding entity block displays the selected related entity field instead of appearing as an empty entry.
* Field counts shown in graph nodes remain synchronized with the configured fields.

AmoghParmar:
### Related Issues

Closes #950

---

This PR fixes an issue in the Data Document Builder where clicking **Add Field** or **Add Related Entity** immediately created an empty field entry before the user selected a field.

Previously, opening the field picker mutated the graph state by creating a placeholder entry with an empty path. If the user selected a field, the selected value was not always persisted into the created entry, leaving an empty field block in the builder. Additionally, cancelling or closing the picker could leave behind unintended blank entries.

This change defers field creation until the user explicitly selects a field. Existing fields continue to be updated in place, while newly added fields are created only after selection. As a result:

* Empty field entries are no longer created when opening the picker.
* Selected fields are persisted correctly.
* Cancelling or closing the picker no longer leaves dangling entries.
* The Data Document Builder state remains consistent until a user confirms a selection.

---

#### Before

Opening **Add Field** or **Add Related Entity** immediately created an empty field entry in the builder.

image

#### After

No entry is created until a field is selected. After selection, the chosen field is added correctly and cancelling the picker leaves the builder unchanged.

image

---

* [x] I read and followed [[contribution rules](https://github.com/hotwax/job-manager#contribution-guideline)](https://github.com/hotwax/job-manager#contribution-guideline)


AmoghParmar:
@dt2patel @ymaheshwari1 @ravilodhi 
Kindly review this PR.

src/views/DataDocumentFormView.vue
```diff
@@ -684,7 +687,7 @@ watch(entityQueryString, () => {
 });
 
 watch([fieldQueryString, modalCurrentEntity], () => {
-  fieldPickerNavigation.resetNavigation();
+  entityPickerNavigation.resetNavigation();
```
dt2patel: @AmoghParmar this watcher belongs to the field picker state. Resetting `entityPickerNavigation` here will leave keyboard navigation stale in the field modal after search/current entity changes. Please switch this back before Yash reviews.

```suggestion
  fieldPickerNavigation.resetNavigation();
```

AmoghParmar: @dt2patel Thanks for reviewing, sir. I switched it back to reset `fieldPickerNavigation` as suggested and pushed the fix.

PR #953, opened 2026-06-18
Fix: Disable already selected fields in Data Document Builder picker (#952)

### LINKED ISSUE #952: Data Document Builder: Disable already selected fields to prevent duplicate selection
## Current behavior

When opening the field picker in the Data Document Builder, users can select the same field multiple times within the same entity context.

Fields that have already been selected remain visible in the picker and can be selected again, which allows duplicate field entries to be created within the same entity context.

## Expected behavior

Fields that have already been added to the current entity context should remain visible in the field picker but should not be selectable again.

Previously selected fields should be displayed in a disabled state so users can clearly identify fields that have already been chosen while still being able to view all available fields.

The implementation can leverage Ionic's `disabled` property on `IonItem`.

Reference:
https://ionicframework.com/docs/api/item#prop-disabled

2. Create a Data Document.

3. Click **Add Field**.

4. Select any field.

5. Click **Add Field** again.

6. Observe the field picker.

### Current Result

The previously selected field remains visible and can be selected again, resulting in duplicate field entries being created.

### Expected Result

The previously selected field should remain visible but should be displayed in a disabled state and cannot be selected again.

* [x] YES
* [ ] NO

## Environment details

* Browser: Chrome 137.x
* OS: Ubuntu 24.04 LTS
* Code Version: `main`

## Additional information

This behavior should be applied consistently for both:

* Add Field
* Add Related Entity

Duplicate field selection must still be prevented.

#### Before Fix

> The screenshot below shows a previously selected field remaining enabled and selectable in the field picker.

Image

> The screenshot below demonstrates that selecting the same field again creates duplicate entries in the Data Document Builder.

Image

AmoghParmar:
Closes #952

### Changes

This PR prevents duplicate field selection in the Data Document Builder. Previously selected fields now remain visible in the picker but are displayed in a disabled state using Ionic's `disabled` property on `IonItem`.

### Root Cause

Previously selected fields remained enabled in the field picker and could be selected multiple times, resulting in duplicate field entries being created within the same entity context.

### Solution

1. **Keep Fields Visible**

   Previously selected fields continue to appear in the picker so users can view all available fields within the current entity context.

2. **Apply Ionic Disabled Prop**

   Bound Ionic's `:disabled` property to `ion-item` elements in the picker modals using helper functions (`isFieldSelected`, `isGraphFieldSelected`, `isRelatedFieldSelected`) to determine whether a field path already exists in the graph.

3. **Prevent Duplicate Selection**

   Added guards to `onSelect` handlers and click events to ensure disabled items cannot be selected through mouse interactions or keyboard navigation.

---

#### Before Fix

The screenshot below demonstrates that selecting the same field multiple times creates duplicate entries in the Data Document Builder.

Image

---

#### After Fix

The screenshot below shows that previously selected fields remain visible in the picker but are displayed in a disabled state and cannot be selected again.

image

---

### Result

* Previously selected fields remain visible in the picker.
* Previously selected fields are displayed in a disabled state.
* Duplicate field selection is prevented.
* The behavior is applied consistently for both **Add Field** and **Add Related Entity** flows.


AmoghParmar:
@dt2patel  @ymaheshwari1  @ravilodhi 
Kindly review this PR.

src/views/DataDocumentGraphBuilder.vue
```diff
@@ -1218,22 +1240,44 @@ const getRelationshipSegments = (fieldPath: string) => {
...
   if (!entityName || !field?.fieldPath) return entityName;
 
   for (const segment of getRelationshipSegments(field.fieldPath)) {
     const relationship = getRelationship(entityName, segment);
-    if (!relationship?.relatedEntityName) return entityName;
-    entityName = relationship.relatedEntityName;
+    if (relationship?.relatedEntityName) {
+      entityName = relationship.relatedEntityName;
+    } else if (segment.includes('#')) {
+      entityName = segment.split('#')[1];
```
dt2patel: @AmoghParmar quick style cleanup: this repo expects double quotes for string literals.

```suggestion
    } else if (segment.includes("#")) {
      entityName = segment.split("#")[1];
```

src/views/DataDocumentGraphBuilder.vue
```diff
@@ -1218,22 +1240,44 @@ const getRelationshipSegments = (fieldPath: string) => {
...
-  ))
-);
+const getRelationship = (entityName: string, relationshipName: string) => {
+  const relations = utilStore.getEntityRelationships(entityName);
+  const lowerName = relationshipName.toLowerCase();
+  
+  const hashIndex = relationshipName.indexOf("#");
+  if (hashIndex > -1) {
+    const title = relationshipName.slice(0, hashIndex).toLowerCase();
+    const nameOrEntity = relationshipName.slice(hashIndex + 1).toLowerCase();
```
dt2patel: @AmoghParmar please avoid `indexOf` here and use the repo-preferred modern string helpers. This keeps the same behavior while removing the style-guide issue.

```suggestion
  if (relationshipName.includes("#")) {
    const [title = "", nameOrEntity = ""] = relationshipName.split("#").map((part) => part.toLowerCase());
```

dt2patel:
@AmoghParmar one more cleanup before Yash reviews: the `DataDocumentGraphBuilder.vue` checkbox handlers should be guarded too. GitHub would not let me attach this as an inline suggestion because those exact lines are not commentable in the diff, but please update these two handlers:

```vue
@ionChange="!isGraphFieldSelected(field.name) && toggleGraphField(field.name, $event.detail.checked)"
```

```vue
@ionChange="!isRelatedFieldSelected(field.fieldName) && toggleRelatedField(field.fieldName, $event.detail.checked)"
```

That keeps mouse/checkbox behavior aligned with the keyboard guard you already added.

AmoghParmar:
@dt2patel Thanks for reviewing, sir. I have updated the checkbox handlers with the selection guards in DataDocumentGraphBuilder.vue, replaced indexOf with includes & split, updated the quote styles to double quotes, and successfully pushed the changes.

ymaheshwari1:
Closing this PR as its linked issue is closed, because we may have cases where the same field needs to be used twice in a data document by adding some function or expression.

PR #971, opened 2026-06-23
fix: require values for applicable condition operators (#957)

### LINKED ISSUE #957: Data Document Builder: Value field should be required when creating conditions
## Current behavior

When creating a condition in the Data Document Builder, the **Value** field is not treated as a required field for operators that expect an explicit value.

Users can create and save a condition even when the **Value** input is left empty.

For example:

| Field Alias | Operator | Value |
|-------------|----------|-------|
| lineItemsHash | Equals | *(empty)* |

The condition is added successfully despite having no value.

---

## Expected behavior

The **Value** field should be required only for operators that expect an explicit value.

Users should not be able to add a condition until a valid value is provided for these operators.

Examples of operators that require a value:

- Equals
- Not equals
- Contains
- Starts with
- In list
- Greater than
- Greater than or equal
- Less than
- Less than or equal
- Between

Operators that do not require an explicit value should remain unaffected.

Examples:

- Is empty
- Is not empty

If the selected operator requires a value and the Value field is empty, validation should prevent condition creation and display an appropriate error message.

Example:

```text
Value is required
```

The **ADD** button should remain disabled until all mandatory fields are populated.

---

2. Create a Data Document.

3. Add a field.

4. Click **ADD +** under **Conditions**.

5. Select an operator that requires a value (for example, **Equals**).

6. Leave the **Value** field empty.

7. Click **ADD**.

### Current Result

The condition is created successfully even though the Value field is empty.

### Expected Result

The condition should not be created.

The **ADD** button should remain disabled or an error message should be displayed until a value is entered.

---

- [x] YES
- [ ] NO

## Environment details

- Browser: Chrome 137.x
- OS: Ubuntu 24.04 LTS
- Code Version: `main`

## Additional information

Validation should only be applied for operators that require an explicit value.

Operators such as **Is empty** and **Is not empty** should continue to work without requiring any value.

### Visual Evidence

The screenshot below demonstrates that the **Value** field is empty while the condition can still be added.

Image

### LINKED ISSUE #970: Data Document Builder: Dismissing condition dialog with X button creates unintended condition
## Current behavior

When creating a condition in the Data Document Builder, dismissing the condition dialog using the **X** button in the top-left corner creates a new condition unexpectedly.

In contrast, clicking the **Cancel** button correctly closes the dialog without creating any condition.

As a result, the dismiss action behaves differently from the Cancel action and may introduce unintended conditions.

---

## Expected behavior

The **X** button should behave identically to the **Cancel** button.

Closing the dialog using either action should:

* Close the dialog
* Discard any unsaved condition changes
* Not create a new condition

---

2. Create a Data Document.
3. Open the **Conditions** dialog.
4. Do not save the condition.
5. Click the **X** button in the top-left corner.

### Visual Evidence

The video below demonstrates the unexpected behavior.

#### Reproduction Flow

1. Initially, there are **no conditions configured** for the Data Document.
2. Click **ADD +** under the **Conditions** section.
3. The condition dialog opens.
4. Without saving the condition, click the **X** button in the top-left corner.
5. Observe that a **new condition is created unexpectedly**.
6. Repeat the same steps and click the **Cancel** button instead.
7. Observe that **no condition is created**, which is the expected behavior.

This demonstrates that the **X** button and the **Cancel** button do not currently perform the same cleanup logic.

#### Video Evidence

https://github.com/user-attachments/assets/8dffe9f3-bb58-4e61-b621-248a051c350d

### Current Result

A new condition is created unexpectedly.

### Expected Result

The dialog should close without creating any condition.

---

## Additional information

This issue does not occur when using the **Cancel** button.

The **Cancel** button already performs the expected cleanup behavior, suggesting that the **X** button does not invoke the same reset/discard logic.

---

## Environment details

* Browser: Chrome 137.x
* OS: Ubuntu 24.04 LTS
* Code Version: `main`

---

* [x] YES
* [ ] NO

AmoghParmar:
### Related Issues

Closes #957
Closes #970

This PR implements mandatory input validation for Data Document Builder conditions, ensuring that operators requiring explicit values cannot be saved with an empty **Value** field.

To improve the user experience, validation feedback is only displayed after a user attempts to add/save the condition rather than immediately when selecting an operator.

During implementation, an additional issue was identified where dismissing the condition dialog using the top-left **X** button could unintentionally create a new condition. This behavior has also been corrected, and both dismissal mechanisms now consistently reset transient validation state.

#### 1. Validation Logic

- Implemented graph-level validation in `projectDataDocumentGraph` to detect missing values for operators that require explicit input and surface a `missing_condition_value` validation issue.
- Added comprehensive unit tests in `dataDocumentGraph.spec.ts` to assert that missing condition values for relevant operators are detected.

#### 2. UI/UX and Validation Visuals

- Introduced a `conditionSubmitted` reactive ref in `DataDocumentGraphBuilder.vue` to track add/save attempts.
- Configured the Condition Dialog to display validation styling and inline error messages only after a user attempts to add/save the condition.
- Ensured validation state resets whenever the modal is opened, closed, dismissed, or swiped away.
- Applied the same validation styling to inline condition editors in `DataDocumentFormView.vue` to maintain a consistent experience across both condition editing interfaces.

#### 3. Additional Dialog Cleanup Fix

- Identified and fixed an unrelated issue where dismissing the condition dialog using the top-left **X** button could unintentionally create a new condition.
- The **Cancel** button and the **X** button now perform identical cleanup behavior and reset transient validation state consistently.

---

#### After Fix

The video below demonstrates the updated validation and dialog dismissal behavior.

It shows that:

- The Condition dialog initially opens without displaying any validation errors.
- The **"Value is required"** message only appears after attempting to add/save a condition with an empty required **Value** field.
- Clicking **Cancel** or the top-left **X** button clears the validation state.
- Dismissing the dialog using the **X** button no longer creates an unintended condition.

[Screencast from 2026-06-23 16-37-57.webm](https://github.com/user-attachments/assets/2c67c0f5-8937-4c72-807d-96435d608682)

---

### Is the changes contains any breaking change?

- [ ] Yes
- [x] No

---

- [x] I read and followed [contribution rules](https://github.com/hotwax/accxui#contribution-guideline)


src/utils/dataDocumentGraph.ts
```diff
@@ -549,7 +559,17 @@ export const projectDataDocumentGraph = ({
         severity: "error",
         message: `Condition references missing field alias "${fieldNameAlias}".`,
         targetKind: "condition",
-        targetId: condition.conditionSeqId
+        targetId: condition.conditionSeqId || condition.localId
+      });
+    }
+
+    if (isConditionValueMissing(graphCondition.operator, graphCondition.fieldValue)) {
```
ymaheshwari1: I think this condition is not required as we are not allowing adding a condition without a value in any case, thus this will never add a validationIssue in the issues tab.

src/views/DataDocumentGraphBuilder.vue
```diff
@@ -760,9 +763,12 @@
               </ion-select>
               <ion-input
                 v-else
-                v-model="activeCondition.fieldValue"
+                :value="activeCondition.fieldValue"
                 :label="translate('Value')"
                 label-placement="stacked"
+                :class="{ 'ion-invalid ion-touched': conditionSubmitted && isOperatorValueInvalid }"
```
ymaheshwari1, line 771: I think we do not need conditionSubmitted variable.

src/views/DataDocumentGraphBuilder.vue
```diff
@@ -744,11 +744,14 @@
             <ion-item>
```
ymaheshwari1, line 746: Add condition to lines property to not display ion-item lines when error is displayed.

ymaheshwari1 (changes requested):

src/store/dataDocumentGraph.spec.ts
```diff
@@ -248,6 +248,8 @@ describe("data document graph projection", () => {
     ]));
   });
 
+
+
```
rachitaarsh-jpg: Unnecessary indentations should be avoided.

src/views/DataDocumentGraphBuilder.vue
```diff
@@ -1740,4 +1760,11 @@ onIonViewWillEnter(async () => {
     min-height: 520px;
   }
 }
+
+ion-item.ion-invalid.ion-touched,
```
ymaheshwari1: We do not need this css.

ymaheshwari1 (changes requested):

PR #992, opened 2026-07-03
Improved: align Message Type Detail actions with header title

### LINKED ISSUE #981: Align Message Type Detail actions with header title
## Problem
On the Message Type Detail page, the Save and Delete actions currently sit below the title/description block, which makes the header feel vertically disconnected and wastes space.

Example page: /system-message-types/AppeasementsFinancialFeed

## Expected behavior
Place the action buttons horizontally aligned with the title area, ideally on the right side of the header, while keeping the title and description readable.

## Notes
- Preserve existing Ionic component patterns.
- Do not introduce broad custom styling; keep the fix scoped to this header layout.
- Ensure the layout still works on mobile, likely by wrapping actions below the title only on narrow widths.

dt2patel:
## Business summary
On the Message Type Detail page the Save and Delete actions sat below the title/description block, leaving the header vertically disconnected and wasting space. The actions are now horizontally aligned with the title, matching the layout the issue requests.

Closes #981

## Changes
- Makes the `.header` block in `SystemMessageTypeDetail.vue` a wrapping flex row: title/description on the left, actions on the right.
- On narrow (mobile) widths the actions wrap below the title, as required by the issue.
- Long unbroken type ids (e.g. `AppeasementsFinancialFeed`) now wrap inside the title instead of overflowing the viewport.

Scoped strictly to this header layout per the issue notes — no broad custom styling, existing Ionic component patterns preserved. `SystemMessageRemoteDetail.vue` shares the same header markup; left untouched since the issue scopes the fix to Message Type Detail, but the same 3 CSS rules can be lifted there if we want parity.

## Testing
- `npm run build` passes.
- Verified in the browser against local Moqui on `/system-message-types/AppeasementsFinancialFeed`: desktop shows title left / actions right on one row; 375px mobile wraps actions below the title and the long id wraps within the viewport.


src/views/SystemMessageTypeDetail.vue
ymaheshwari1, line 1: This changes needs to be done on the SystemMessageRemoteDetail page as well.

Also we can align button to center instead of having them at start.

ymaheshwari1 (changes requested):

src/views/SystemMessageTypeDetail.vue
```diff
@@ -281,6 +281,20 @@ onIonViewWillEnter(() => {
 </script>
 
 <style scoped>
+.header {
+  display: flex;
+  align-items: flex-start;
+  justify-content: space-between;
+  flex-wrap: wrap;
+  gap: 12px;
```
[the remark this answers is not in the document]
dt2patel: Disagree — leaving this styling as-is. This is a styling-only change, and this PR should not touch existing scoped styling code unless there is a functional issue to fix.

src/views/SystemMessageTypeDetail.vue
dt2patel: Agree — fixed in 34b4b4cb by applying the same header layout to `SystemMessageRemoteDetail` and centering the header action buttons.

PR #1020, opened 2026-07-09
Fix average speed calculation for Dashboard KPI cards

### LINKED ISSUE #1016: Bug: Dashboard KPI cards show incorrect Avg. Speed due to mismatched timestamp field names
## Current behavior

On the Dashboard, the **Avg. Speed** display for three KPI cards is always 
incorrect:

- **High-priority ingestion** → shows `--` (no average computed)
- **Standard ingestion** → shows `--` (no average computed)  
- **Incoming messages** → shows `0s` (zero or near-zero average)

This is the case even when there are multiple **Finished** ingestion logs or 
**Consumed** messages in the dataset that should produce a meaningful average.

## Expected behavior

The Avg. Speed for each card should display the real average processing time 
calculated from the actual start/end timestamps of finished/consumed records 
in the latest fetched batch.

1. Open the **Dashboard** page with a populated dataset (finished ingestion 
   logs and consumed messages present).
2. Observe the **High-priority ingestion** card → Avg. Speed shows `--`.
3. Observe the **Standard ingestion** card → Avg. Speed shows `--`.
4. Observe the **Incoming messages** card → Avg. Speed shows `0s`.

## Root cause

**High-Priority & Standard (--):** `getAvgProcessingTime()` uses incorrect API field names, causing the calculated duration to be `0`, which is displayed as `--`.

**Incoming Messages (0s):** `getAvgMessageProcessingTime()` skips messages without `processedDate`. If none of the fetched messages are processed, it returns `0`, which is displayed as `0s` instead of `--`.

- [x] YES

## Environment details

- Browser: Chrome (latest)
- OS: Linux
- Code Version: Current `main` branch

rachitaarsh-jpg:
### Related Issues

Closes #1016 

This PR fixes the average speed calculations for the KPI cards on the Job Manager Dashboard. 
Previously, the Avg. Speed for High-priority and Standard ingestion was displaying as `--`, and Incoming messages was displaying as `0s`.

**Changes made:**
- **Ingestion Cards:** Corrected the end-time field mapping in `getAvgProcessingTime()` from `lastUpdatedStamp` (which was returning undefined/null) to `finishDateTime || lastUpdatedTxStamp` to properly calculate the duration.
- **Messages Card:** Fixed the `getAvgMessageProcessingTime()` calculation to first filter out messages that do not have a `processedDate`. Previously, these unmeasurable messages were being included in the denominator, artificially skewing the average down to 0.

This is useful because it ensures the Dashboard displays accurate and reliable performance metrics for system health monitoring.

### DIFF
--- a/src/views/Pipeline.vue
+++ b/src/views/Pipeline.vue
@@ -720,7 +720,7 @@ const getAvgProcessingTime = (logsList: any[]) => {
   if (finishedLogs.length === 0) return 0;
   const totalSeconds = finishedLogs.reduce((acc, log) => {
     const start = log.createdDate;
-    const end = log.lastUpdatedStamp;
+    const end = log.finishDateTime || log.lastUpdatedTxStamp;
     if (!start || !end) return acc;
     const startDt = typeof start === 'number' ? DateTime.fromMillis(start) : DateTime.fromISO(start);
     const endDt = typeof end === 'number' ? DateTime.fromMillis(end) : DateTime.fromISO(end);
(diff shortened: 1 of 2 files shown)


src/views/Pipeline.vue
```diff
@@ -720,7 +720,7 @@ const getAvgProcessingTime = (logsList: any[]) => {
   if (finishedLogs.length === 0) return 0;
   const totalSeconds = finishedLogs.reduce((acc, log) => {
     const start = log.createdDate;
-    const end = log.lastUpdatedStamp;
+    const end = log.finishDateTime || log.lastUpdatedTxStamp;
```
aniket-date: Use this instead.
```suggestion
    const end = log.finishDateTime || log.lastUpdatedStamp;
```

rachitaarsh-jpg: Sir, I had used lastUpdatedTxStamp to keep the logic consistent with how we are currently calculating durations in FileHistory.vue. I'll update it to lastUpdatedStamp here as suggested, and I will also push an update to FileHistory.vue in this PR so that both pages remain in sync.

rachitaarsh-jpg:
### **_Note- TestScript is in making_**

src/views/Pipeline.vue
```diff
@@ -772,19 +772,21 @@ const successMessagesCount = computed(() => systemMessages.value.filter((msg: an
 const getAvgMessageProcessingTime = (messagesList: any[]) => {
   const finishedMessages = messagesList.filter((msg: any) => ['SmsgSent', 'SmsgConsumed', 'SmsgConfirmed'].includes(msg.statusId));
   if (finishedMessages.length === 0) return 0;
-  const totalSeconds = finishedMessages.reduce((acc, msg) => {
+  const measurableMessages = finishedMessages.filter((msg: any) => msg.initDate && msg.processedDate);
+  if (measurableMessages.length === 0) return 0;
+  const totalSeconds = measurableMessages.reduce((acc, msg) => {
```
ymaheshwari1: I think this changes are not required.

ymaheshwari1 (changes requested):

PR #1021, opened 2026-07-09
Added: support to filter messages by inbound/outbound and improved pagination UI

### LINKED ISSUE #1014: Bug: Priority filter not applied when navigating to File History from Dashboard
## Summary

Navigating from the Dashboard to File History via the High-priority or Standard ingestion cards passes `?priority=HIGH` (or `?priority=NORMAL`) in the URL, but the File History page ignores this query parameter on load. As a result, 
the page opens with no priority filter applied, showing all jobs instead of only the expected priority group — even though the URL correctly reflects the intended filter. Manually selecting the same filter from the dropdown works fine.

Additionally, the same category of bug affects the Queue Operations Map section on the Dashboard — both for File History (priority) links and Message History (inbound/outbound direction) links. In all cases, the destination page ignores the query parameters passed in the URL on load.

---

## Bug Report

**Affected Pages:** Dashboard → File History, Dashboard → Message History  
**Files:**  
- `src/views/FileHistory.vue`  
- `src/views/SystemMessageMonitor.vue`

---

## Steps to Reproduce

**Case 1 — High-priority / Standard ingestion cards (KPI section)**
1. Open the **Dashboard** page.
2. Click on any chip under **"High-priority ingestion"** or **"Standard ingestion"**.
3. Observe the File History page — URL shows `?priority=HIGH` (or `NORMAL`).
4. Observe the Priority filter dropdown — it shows **"All"** (unfiltered).
5. All jobs are listed instead of only the priority-filtered ones.

**Case 2 — Queue Operations Map → High-priority Queue / Standard Queue**
1. Open the **Dashboard** page.
2. Click **"High-priority queue"** or **"Standard queue"** in the Queue Operations 
   Map section.
3. Same result as Case 1 — priority filter is not applied on the File History page.

**Case 3 — Queue Operations Map → Inbound Queue / Outbound Queue**
1. Open the **Dashboard** page.
2. Click **"Inbound queue"** — navigates to `/system-messages?isOutgoing=N`.
3. Click **"Outbound queue"** — navigates to `/system-messages?isOutgoing=Y`.
4. Observe the Message History page — the list shows **all messages mixed** 
   (both inbound and outbound), with no direction filter applied.
5. Manually applying the filter after landing works correctly.

---

## Expected Behavior

- `/file-history?priority=HIGH` → Pre-applies **High** priority filter; shows 
  only high-priority jobs.
- `/file-history?priority=NORMAL` → Pre-applies **Normal** priority filter.
- `/system-messages?isOutgoing=N` → Pre-applies inbound filter; shows only 
  inbound (incoming) messages.
- `/system-messages?isOutgoing=Y` → Pre-applies outbound filter; shows only 
  outbound (outgoing) messages.

---

## Actual Behavior

- `FileHistory.vue` — `onIonViewWillEnter` only reads `statusId` from the 
  route query and ignores `priority`. Priority filter remains **"All"**.
- `SystemMessageMonitor.vue` — `onIonViewWillEnter` only reads `statusId` 
  from the route query and ignores `isOutgoing`. Both inbound and outbound 
  messages are shown together with no direction filter applied.

In all cases, manually selecting the filter after landing works correctly, 
confirming the filter logic itself is sound — it is simply never initialized 
from the URL on navigation.

---

rachitaarsh-jpg:
### Related Issues

Closes  #1014 

This PR resolves cross-page navigation filtering issues between the Job Manager Dashboard and the File/Message History pages, and unifies the UI and pagination logic. 

**Key Changes:**

* **Dashboard Navigation Fix:** Ensures query parameters (e.g., `?priority=HIGH` or `?isOutgoing=N`) passed from Dashboard KPI cards correctly initialize and apply filters in `onIonViewWillEnter` upon loading the target history pages. Solved race conditions by moving state assignments after initial metadata fetches.

* **Hybrid Pagination for Client-Side Filtering:** Since `isOutgoing` is not supported as a server-side filter, the System Message Monitor now fetches a larger batch (500 records) when a direction filter is active. It performs client-side filtering and pagination locally against this batch to ensure page counts are accurate and prevent empty pages. 

* **Unified UI & UX:** Updated the System Message Monitor filter grid to match the File History page:
    * Implemented a responsive single-row flex grid for all filter dropdowns.
    * Added conditional "Clear" (close circle) buttons that appear only when a filter is active.
    * Replaced static pagination text with an editable page number input field, allowing users to jump directly to specific pages.

### DIFF
--- a/src/views/FileHistory.vue
+++ b/src/views/FileHistory.vue
@@ -576,10 +576,10 @@ onIonViewWillEnter(async () => {
   } else {
     await mdmStore.updateAppliedFilters("statusId", []);
   }
-  if(currentQuery?.priority){
+  if (currentQuery?.priority) {
     selectedPriority.value = [(currentQuery.priority as string)];
-  }else{
-      selectedPriority.value = [];
+  } else {
+    selectedPriority.value = [];
   }
   await fetchLogs();
   mdmStore.fetchConfigs();
--- a/src/views/FileHistory.vue
+++ b/src/views/FileHistory.vue
@@ -569,10 +569,17 @@ watch(pageIndex, () => {
 });
 
 onIonViewWillEnter(async () => {
-  if (route.query?.statusId) {
-    await mdmStore.updateAppliedFilters("statusId", (route.query.statusId as string).split(","));
+  const currentQuery = router.currentRoute.value.query;
+
+  if (currentQuery?.statusId) {
+    await mdmStore.updateAppliedFilters("statusId", (currentQuery.statusId as string).split(","));
   } else {
     await mdmStore.updateAppliedFilters("statusId", []);
+  }
+  if(currentQuery?.priority){
+    selectedPriority.value = [(currentQuery.priority as string)];
+  }else{
+      selectedPriority.value = [];
   }
   await fetchLogs();
   mdmStore.fetchConfigs();
--- a/src/views/Pipeline.vue
+++ b/src/views/Pipeline.vue
@@ -199,7 +199,7 @@
                         <ion-icon slot="start" :icon="cloudUploadOutline" color="secondary" />
                         <ion-label>
                           {{ translate("High-Priority Queue") }}
-                          <p>{{ highPriorityPendingCount }} {{ translate("Pending Files") }}</p>
+                          <p>{{ mapHighPriorityCount }} {{ translate("Total Files") }}</p>
                         </ion-label>
                       </ion-item>
                       <div class="arrow-container">
@@ -209,7 +209,7 @@
                         <ion-icon slot="start" :icon="cloudUploadOutline" color="medium" />
                         <ion-label>
                           {{ translate("Standard Queue") }}
-                          <p>{{ standardPendingCount }} {{ translate("Pending Files") }}</p>
+                          <p>{{ mapStandardCount }} {{ translate("Total Files") }}</p>
                         </ion-label>
                       </ion-item>
                     </div>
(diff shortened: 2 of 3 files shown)


rachitaarsh-jpg:
### **_Note- TestScript is in making_**

src/views/FileHistory.vue
```diff
@@ -569,11 +569,18 @@ watch(pageIndex, () => {
...
+  const currentQuery = router.currentRoute.value.query;
+
+  if (currentQuery?.statusId) {
+    await mdmStore.updateAppliedFilters("statusId", (currentQuery.statusId as string).split(","));
   } else {
     await mdmStore.updateAppliedFilters("statusId", []);
   }
+  if (currentQuery?.priority) {
+    selectedPriority.value = [(currentQuery.priority as string)];
+  } else {
```
ymaheshwari1: Is else part required?

src/views/FileHistory.vue
```diff
@@ -569,11 +569,18 @@ watch(pageIndex, () => {
 });
 
 onIonViewWillEnter(async () => {
-  if (route.query?.statusId) {
```
ymaheshwari1: This should work fine, and we can directly use route.query instead of having another new variable as currentQuery

ymaheshwari1 (changes requested):
I think this PR should only focus on the navigation issue and it should not include the UI optimization or other pagination fixes, update this PR for navigation changes and create separate PR for other fixes.

PR #1022, opened 2026-07-09
fix: File History status filter and KPI count

### LINKED ISSUE #1013: Fix File History status filter for "Finished with errors" records
**Problem**
The Status filter in MDM → File History does not return files that contain failed executions.
The status displayed for these files is "Finished with errors", but the filter only checks for the "Failed" status. As a result, applying the Failed filter returns no results, even though files with failed executions exist.

**Root Cause**
The filter logic is based on the "Failed" status, whereas the actual status stored/displayed for these executions is "Finished with errors".

**Fix**
Updated the status filter logic so that files with the status "Finished with errors" are included when the Failed status filter is applied.

**Expected Behavior**
When the Failed status filter is selected, it should return files whose execution status is "Finished with errors".

harshmahajan7:
# Fix: Failed Status Filter in MDM → File History

## Problem

The **Status** filter in **MDM → File History** does not return files
that contain failed executions.

The status displayed for these files is **"Finished with errors"**, but
the filter only checks for the **"Failed"** status. As a result,
applying the **Failed** filter returns no results, even though files
with failed executions exist.

## Root Cause

The filter logic was only checking for the database statuses:

-   `DmlsFailed`
-   `DmlsCrashed`

However, **"Finished with errors"** is a UI-level status where:

-   Database status = `DmlsFinished`
-   `failedRecordCount > 0`

Since these records were excluded from the query, they never appeared in
the Failed filter.

### 1. Filter Expansion

When the **Failed** status filter is selected, the client now requests
both:

-   `DmlsFailed`
-   `DmlsFinished`

records from the server.

### 2. Client-side Filtering

After receiving the data, the client filters `DmlsFinished` records and
keeps only those where:

``` text
failedRecordCount > 0
```

These records are displayed as **"Finished with errors"**.

## Expected Behavior

When the **Failed** status filter is selected, it correctly returns:

-   Standard failed files (`DmlsFailed`)
-   Files with **"Finished with errors"** (`DmlsFinished` with
    `failedRecordCount > 0`)

The File History list and KPI counts now accurately represent all failed
executions.

### DIFF
--- a/src/store/mdmConfig.ts
+++ b/src/store/mdmConfig.ts
@@ -139,18 +139,63 @@ export const useMdmConfigStore = defineStore("mdmConfig", {
           payload["statusId_op"] = "in"
         }
 
-        const resp = await api({
-          url: "admin/dataManager/details",
-          method: "get",
-          params: payload
-        })
+        const statusIds = payload.statusId ? payload.statusId.split(",") : [];
+        if (statusIds.includes("DmlsFailed") && !statusIds.includes("DmlsFinished")) {
+          // Splitting queries to fetch only explicitly failed/crashed logs and finished logs with errors
+          const statusList1 = [...new Set(statusIds.flatMap((s: string) => s === "DmlsFailed" ? ["DmlsFailed", "DmlsCrashed"] : [s]))].filter(s => s !== "DmlsFinished");
 
-        if (resp.data?.dataManagerLogsCount) {
-          this.logs = resp.data.dataManagerLogs
-          this.logsCount = resp.data.dataManagerLogsCount
+          const payload1 = { ...payload };
+          if (statusList1.length > 0) {
+            payload1.statusId = statusList1.join(",");
+            payload1.statusId_op = "in";
+          } else {
+            delete payload1.statusId;
+            delete payload1.statusId_op;
+          }
+
+          const payload2 = {
+            ...payload,
+            statusId: "DmlsFinished"
+          };
+          delete payload2.statusId_op;
+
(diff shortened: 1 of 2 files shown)


src/store/mdmConfig.ts
```diff
@@ -230,17 +230,21 @@ export const useMdmConfigStore = defineStore("mdmConfig", {
...
           api({ url: "admin/dataManager/details", method: "get", params: { pageSize: 1000, pageIndex: 0, statusId: moquiStatuses, statusId_op: "in" } })
         ])
+        const allLogs = logsResp.data?.dataManagerLogs || []
         this.globalStats = {
-          total: totalResp.data?.dataManagerLogsCount || 0,
-          successful: successResp.data?.dataManagerLogsCount || 0,
-          failed: failedResp.data?.dataManagerLogsCount || 0,
-          avgProcessingTime: getAverageProcessingTime(logsResp.data?.dataManagerLogs || [])
+          total: allLogs.length,
+          successful: allLogs.filter((log: any) =>
```
ymaheshwari1: I think this is not a good approach, as we have a cap of 1000 pageSize and what if all the 1000 records are success, then in that case we do not get the correct ratio for success and failed records.

ymaheshwari1 (changes requested):

ymaheshwari1 (changes requested):
This might result in an empty page in specific case where the failed status filter is applied and we have passed pageSize in api, now in this case we will fetch logs for both the failed and finished status and suppose on first page all the finished status logs are returned with no errors, then the UI might display empty state even when we have actual failed records on next page or finished with error records.

PR #1019, opened 2026-07-09
Fixed: check for empty values when updating remote and type details

ymaheshwari1:
### Related Issues

#

### Contribution and Currently Important Rules Acceptance

- [x] I read and followed [contribution rules](https://github.com/hotwax/job-manager#contribution-guideline)

PR #1025, opened 2026-07-11
Fix credential wiping when updating Remote System Details

### LINKED ISSUE #1012: Bug: Saving Remote System Detail silently clears stored credentials from the database
## Summary

Saving any field on the Remote System Detail page silently clears existing credential fields (Password, Private Key, Shared Secret, etc.) from the database. When a user edits a non-sensitive field such as Description and clicks Save, the credentials are overwritten with empty strings in the `system_message_remote` table — with no warning shown to the user. This was reproduced on `AWS_CONFIG` where the password `gorjana_aws_config` was wiped after only updating the 
Description field.

---

## Bug Report

**Affected Page:** Remote System Detail (`/system-message-remotes/:remoteId`)  
**File:** `src/views/SystemMessageRemoteDetail.vue`  

---

## Steps to Reproduce

1. Open an existing Remote System (e.g., `AWS_CONFIG`) that has a Password set.
2. Edit only the **Description** field.
3. Click **Save**.
4. Verify the `system_message_remote` table in the database — the Password 
   column is now empty/null.

---

## Expected Behavior

Only the explicitly edited field (Description) should be updated.  
Credential fields (**Password**, **Private Key**, **Shared Secret**, 
**Send Shared Secret**, **Old Shared Secret**) that were not touched by 
the user should remain unchanged in the database.

---

## Actual Behavior

All credential fields are sent as empty strings (`""`) in the save payload, 
overwriting their existing values in the database — even though the user 
never interacted with them.

---

rachitaarsh-jpg:
### Related Issues

Closes #1012 

**Fix: Prevent credential wiping when updating Remote System Details**

Previously, updating any non-sensitive field (like Description) on the Remote System Detail page would silently overwrite all credential fields (Password, Private Key, Shared Secret, etc.) with empty strings in the database. This happened because the backend deliberately (and correctly) omits these sensitive fields from the GET response, but the frontend was defaulting undefined fields to `""` and including them in the PUT payload.

This PR fixes the issue by:
1. Defining a `CREDENTIAL_FIELDS` set.
2. Initializing untouched credential fields as `null` instead of `""`.
3. Filtering out `null` credential fields when constructing the save payload.

As a result, updating non-sensitive fields no longer affects existing credentials in the database, while explicit password changes by the user still work as expected.

### DIFF
--- a/src/views/SystemMessageRemoteDetail.vue
+++ b/src/views/SystemMessageRemoteDetail.vue
@@ -179,14 +179,6 @@ const form = reactive<Record<string, any>>({
   "sendServiceName": { label: "Send Service", type: "textarea" }
 });
 
-const CREDENTIAL_FIELDS = new Set([
-  "password",
-  "privateKey",
-  "sharedSecret",
-  "sendSharedSecret",
-  "oldSharedSecret"
-]);
-
 const isCreateMode = computed(() => !props.id);
 const pageTitle = computed(() => isCreateMode.value ? translate("Create Remote System") : translate("Remote System Detail"));
 const statuses = computed(() => utilStore.getStatusItemsByType("SystemMessage"));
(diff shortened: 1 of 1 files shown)


ymaheshwari1:
@rachitaarsh-jpg Please check this PR https://github.com/hotwax/job-manager/pull/1019/changes

Banibrata-Manna (approved):

PR #1029, opened 2026-07-14
fix generic system-message-remote-id jobs not showing

rachitaarsh-jpg:
### Related Issues

Closes 1009 

**Refactor Job Filtering Logic and Remove Hardcoded Remote Identifiers**

This PR introduces two major improvements to how the Job Manager filters and displays jobs:

1. **Environment-Driven Shared Remotes:** Replaced the hardcoded `.toLowerCase().includes("sftp")` bypass logic with a dynamic, environment-driven approach. The app now uses the `VITE_SHARED_REMOTE_IDENTIFIERS` environment variable to determine which system message remote IDs should be treated as "shared" (global) connections. This decoupling prevents store-specific jobs from leaking across views while safely allowing true shared connections (e.g., `RemoteSftp`, `AWS_CONFIG`) to bypass the store dependency filter via a flexible substring match.
2. **DRY Refactoring:** Extracted duplicate dependency filtering logic found in both `fetchJobs` and `fetchJobDetail` into a single, reusable `evaluateJobDependency(job)` helper function. This significantly cleans up the codebase and ensures that any future changes to job evaluation rules only need to be applied in one place.


rachitaarsh-jpg:
Closing this, another approach will be used to fix this issue.

PR #1035, opened 2026-07-21
Fix: Improve Recent Exports helper text spacing and alignment (#1034)

### LINKED ISSUE #1034: Bug: Recent Exports helper text overlaps the top border in Data Document creation flow
### Current behavior

When creating a new **Data Document** in Job Manager and navigating to the **Recent Exports** tab, the helper text at the top of the panel is positioned too close to the container border.

The text:

> "Exports run the full document (with its conditions) and include up to 10,000 rows."

touches the top edge of the panel, making the UI appear cramped and reducing readability.

---

### Steps to reproduce

1. Open **Job Manager**.
2. Navigate to **Data Documents**.
3. Click **Create**.
4. Open the **Recent Exports** tab (bottom-right section).
5. Observe the helper text displayed at the top of the panel.

---

### Current result

- The helper text is positioned against the top border of the container.
- There is insufficient top spacing/padding, causing the text to appear visually cramped.

---

### Expected result

- The helper text should have adequate spacing from the top border.
- The panel should maintain consistent padding and alignment with the rest of the UI.
- The content should be clearly readable without appearing to touch the container boundary.

---

### Impact

- Reduces visual clarity.
- Creates an inconsistent UI appearance.
- Makes the Recent Exports section look misaligned compared to other panels.

---

### Screenshot

Image

The attached screenshot demonstrates the helper text touching the top border of the **Recent Exports** panel.

AmoghParmar:
### Bug Fix Summary

Fixes a UI spacing issue in the **Recent Exports** panel within the Data Document creation flow (`DataDocumentGraphBuilder.vue`).

### Changes Made

- Wrapped the helper text inside an `ion-list` to provide consistent container spacing.
- Updated the helper text component to use `ion-label` with `color="medium"` for consistency with the rest of the application.
- Applied `ion-text-center` and `ion-text-wrap` classes to improve alignment and readability.
- Adjusted the layout so the helper text no longer appears flush against the top border of the panel.

### Modified Files

- `src/views/DataDocumentGraphBuilder.vue`

### Verification

- Verified the helper text has consistent spacing from the top border.
- Verified the text is centered and wraps correctly.
- Verified no visual regressions in other tabs or panels.
- Confirmed the Recent Exports panel maintains a consistent appearance with the rest of the application.

### Screenshot
image

Fixes #1034

### DIFF
--- a/src/views/DataDocumentGraphBuilder.vue
+++ b/src/views/DataDocumentGraphBuilder.vue
@@ -417,13 +417,11 @@
           </ion-list>
 
           <template v-else>
-            <ion-list>
-              <ion-item lines="none">
-                <ion-label class="ion-text-center ion-text-wrap" color="medium">
-                  {{ translate("Exports run the full document (with its conditions) and include up to 10,000 rows.") }}
-                </ion-label>
-              </ion-item>
-            </ion-list>
+            <ion-item lines="none" class="ion-text-center">
+              <ion-note class="ion-padding-vertical ion-text-wrap" color="medium" style="width: 100%; font-size: 14px; text-align: center;">
+                {{ translate("Exports run the full document (with its conditions) and include up to 10,000 rows.") }}
+              </ion-note>
+            </ion-item>
             <DataDocumentExportList :messages="exportHistory" :empty-message="translate('No recent exports.')" />
             <ion-list>
               <ion-item button @click="router.push('/data-document-export-history')">
--- a/src/views/DataDocumentGraphBuilder.vue
+++ b/src/views/DataDocumentGraphBuilder.vue
@@ -418,7 +418,7 @@
 
           <template v-else>
             <ion-item lines="none" class="ion-text-center">
-              <ion-note class="ion-padding-vertical ion-text-wrap" color="medium" style="width: 100%; font-size: 14px; text-align: center;">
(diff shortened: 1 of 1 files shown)


src/views/DataDocumentGraphBuilder.vue
```diff
@@ -417,9 +417,13 @@
           </ion-list>
 
           <template v-else>
-            <ion-item lines="none">
-              <ion-note class="ion-text-wrap">{{ translate("Exports run the full document (with its conditions) and include up to 10,000 rows.") }}</ion-note>
-            </ion-item>
+            <ion-list>
+              <ion-item lines="none">
+                <ion-label class="ion-text-center ion-text-wrap" color="medium">
+                  {{ translate("Exports run the full document (with its conditions) and include up to 10,000 rows.") }}
```
ymaheshwari1: This is helper info text, hence its good to be in ion-note, also I don't think the text is being cut on the top, its just too close and I don't think its causing any specific problem here, but if we want to improve it a bit then instead of wrapping it in ion-list, we can simply add ion-padding-vertical class to ion-note and it will solve the problem.

@AmoghParmar

ymaheshwari1 (changes requested):

PR #1041, opened 2026-07-21
fix: restore navigation filters from URL query params on view enter

### LINKED ISSUE #1014: Bug: Priority filter not applied when navigating to File History from Dashboard

rachitaarsh-jpg:
### Related Issues
Closes #1014 

**Short Description:** 
Fixed an issue where filters applied via URL parameters (such as `priority`, `isOutgoing`, and `statusId`) were not being correctly applied when navigating to the File History and System Message Monitor views. 

**Why It's Useful:**
Previously, the views were using a stale snapshot of `route.query` captured during component setup, meaning subsequent navigations back to the view with different parameters were ignored. By switching to `router.currentRoute.value.query` inside `onIonViewWillEnter`, the app now correctly reads the live query state. Additionally, this PR adds the missing logic to extract and apply the `priority` filter in `FileHistory` and the `isOutgoing` client-side filter in `SystemMessageMonitor` directly from the URL.

N/A - Logical and navigation state changes only.

### DIFF
--- a/src/views/SystemMessageMonitor.vue
+++ b/src/views/SystemMessageMonitor.vue
@@ -86,18 +86,6 @@
                   {{ remote.description || remote.systemMessageRemoteId }}
                 </ion-select-option>
               </ion-select>
-
-              <ion-select
-                :label="translate('Direction')"
-                label-placement="stacked"
-                interface="popover"
-                :value="selectedIsOutgoing"
-                @ionChange="selectedIsOutgoing = $event.detail.value"
-              >
-                <ion-select-option value="">{{ translate("All") }}</ion-select-option>
-                <ion-select-option value="N">{{ translate("Inbound") }}</ion-select-option>
-                <ion-select-option value="Y">{{ translate("Outbound") }}</ion-select-option>
-              </ion-select>
             </div>
           </ion-card-content>
         </ion-card>
@@ -148,7 +136,6 @@ const PAGE_SIZE = 25;
 
 const store = useSystemMessageStore();
 const utilStore = useUtilStore();
-const route = router.currentRoute.value;
 
 const queryString = ref("");
 const selectedStatusId = ref("");
--- a/src/views/SystemMessageMonitor.vue
+++ b/src/views/SystemMessageMonitor.vue
@@ -181,8 +181,8 @@ const loadMessages = async () => {
   // (the API does not support isOutgoing as a server-side filter)
   const isDirectionFiltered = !!selectedIsOutgoing.value;
   const payload = {
-    pageIndex: isDirectionFiltered ? 0 : pageIndex.value,
-    pageSize: isDirectionFiltered ? 500 : PAGE_SIZE,
+    pageIndex: pageIndex.value,
+    pageSize: PAGE_SIZE,
   } as Record<string, any>;
 
   if(queryString.value.trim()) {
--- a/src/views/SystemMessageMonitor.vue
+++ b/src/views/SystemMessageMonitor.vue
@@ -242,22 +242,6 @@ const goToNextPage = () => {
   pageIndex.value += 1;
 };
 
-const validatePageInput = (event: any) => {
-  const value = parseInt(event.target.value);
-  if (value > pageCount.value) {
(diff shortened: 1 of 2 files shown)


src/views/SystemMessageMonitor.vue
```diff
@@ -136,33 +136,54 @@ const PAGE_SIZE = 25;
...
 const selectedStatusId = ref("");
 const selectedTypeId = ref("");
 const selectedParentTypeId = ref("");
 const selectedRemoteId = ref("");
+const selectedIsOutgoing = ref("");
 const pageIndex = ref(0);
 
-const messages = computed(() => store.getSystemMessages);
+// When direction filter is active, all fetched records are filtered client-side
+const filteredMessages = computed(() => {
```
ymaheshwari1: As discussed we will add filtering support on backend, so adding filtering based on isOutgoing is not required here.

ymaheshwari1 (changes requested):

src/views/SystemMessageMonitor.vue
rachitaarsh-jpg: Okay Sir.

PR #1058, opened 2026-07-31
Fixed: Prevent File Details crashes on large payloads

### LINKED ISSUE #1056: Prevent Data Manager File Details from crashing on large payloads
## Current behavior

The **Data Manager File Details** payload viewer becomes unresponsive or crashes on larger CSV/JSON/text files. The issue is caused by the client processing and rendering the entire payload eagerly on the browser main thread. It is not caused by invalid data.

Searching also changes the content height substantially, causing layout shift. The payload viewport must have a stable fixed height.

## Reproduction

1. Open Job Manager → File history.
2. Open a log whose original or error payload is large (for example, a CSV around 250,000 rows / 29 MB).
3. Observe the page while the payload is downloaded, parsed, made reactive, and rendered.
4. Type into **Filter rows** / **Search keys and values**.
5. Switch between Original and Errors when both payloads exist.

Typical results: long main-thread stalls, tab memory growth, slow search, large layout shifts, or a browser tab crash.

## Controlled scale evidence

A local benchmark using Papa Parse 5.5.3 with the same `header: true, skipEmptyLines: true` configuration parsed a synthetic 29,111,159-byte CSV with 250,000 rows and 8 columns:

* Parsed rows: 250,000
* Potential table cells: 2,000,000
* Heap increase during parse: approximately 167 MB
* One full-row substring filter pass: approximately 88 ms

This benchmark ran in Node and excludes Vue reactivity, browser layout/paint, and DOM memory, so it is a lower-bound diagnostic rather than a browser performance claim.

## Expected

* Opening File Details remains responsive for large payloads.
* Parsing and search do not block the main thread.
* DOM size is bounded independently of file row count.
* Only the active payload is fetched/parsed unless the user asks for the other tab.
* The payload viewport keeps a stable fixed height while searching, switching formats, loading, showing no matches, or displaying errors.
* Download remains available even when a file is too large for an unrestricted preview.

## Actual code path

1. `loadPayloads` downloads the original and error files concurrently, even though only Original is initially visible.
2. `fetchDataManagerFileContent` materializes the entire response as a JavaScript string.
3. `detectAndParse` creates another trimmed string, then runs synchronous `JSON.parse` or Papa Parse.
4. The async wrapper does not offload work: the shared `parseCsv` calls `Papa.parse` without `worker: true`.
5. State retains both `rawText` and the parsed object/row array for both payload tabs.
6. CSV renders every row and cell with nested `v-for`; JSON starts expanded and recursively mounts all nodes.
7. CSV search scans every value in every row for each debounced query.
8. JSON search calls recursive `deepMatch` in every node, repeating subtree traversal.
9. CSV/JSON use only `max-height: 70vh`; text is vertically unbounded. Filtered results and empty states shrink the viewer, producing layout shift.

Source:

* [FileHistoryDetail.vue](https://github.com/hotwax/job-manager/blob/4722705eee3e6eb2632338d2fe82540ba98bcdab/src/views/FileHistoryDetail.vue#L117-L175)
* [FileHistoryDetail parsing path](https://github.com/hotwax/job-manager/blob/4722705eee3e6eb2632338d2fe82540ba98bcdab/src/views/FileHistoryDetail.vue#L245-L415)
* [JsonNode.vue](https://github.com/hotwax/job-manager/blob/4722705eee3e6eb2632338d2fe82540ba98bcdab/src/components/JsonNode.vue#L1-L111)
* [mdmConfig.ts](https://github.com/hotwax/job-manager/blob/4722705eee3e6eb2632338d2fe82540ba98bcdab/src/store/mdmConfig.ts#L159-L179)
* [shared parseCsv on current AccxUI main](https://github.com/hotwax/accxui/blob/e20f155d999966ab903093ffc565344dc262bf68/common/utils/commonUtil.ts#L745-L760)
* App `main` evidence SHA: `4722705eee3e6eb2632338d2fe82540ba98bcdab`
* Shared AccxUI `main` evidence SHA: `e20f155d999966ab903093ffc565344dc262bf68` (2026-07-30)

## Probable root cause

The viewer architecture assumes the complete file can be represented simultaneously as raw text, parsed reactive data, and a full DOM tree. Promises make the flow asynchronous but do not move CPU work off the UI thread. Memory and DOM work therefore scale with the whole file; when both Original and Errors exist, much of that cost is duplicated.

## Recommended implementation

* Fetch metadata first and lazy-load only the selected payload tab.
* Parse CSV and JSON in a dedicated Web Worker (Comlink is already available in the shared workspace, or use Papa Parse worker/chunk support for CSV).
* Use streaming/chunked parsing and return bounded pages/chunks rather than one massive reactive array.
* Virtualize the CSV rows and JSON nodes so mounted DOM remains bounded.
* Default large JSON trees to collapsed/lazy children.
* Run search/indexing in the worker, cancel superseded searches, and return matching row/node identifiers or bounded result pages.
* Avoid retaining both raw and parsed copies unless required. Prefer `shallowRef`/`markRaw` for large immutable worker results and release inactive payload state.
* Provide a size-aware preview policy: for files above the supported preview threshold, show a bounded preview with an explicit “preview truncated” state while preserving direct download.
* Use one stable viewer shell for CSV, JSON, text, loading, empty, and error states. Recommended CSS shape: an explicit responsive height (for example `height: clamp(360px, 70vh, 720px)`) with `overflow: auto`, not content-driven `max-height`.
* Terminate workers/abort fetches when the route/log ID changes or the view is left.
* Do not make Copy attempt an unrestricted full-file clipboard write above the safe preview threshold; offer Download instead.

## Acceptance criteria

- [ ] A representative 30 MB / 250,000-row CSV opens without a tab crash and the page controls remain responsive.
- [ ] Parsing and full-dataset search execute off the browser main thread; Chrome Performance shows no parse/filter long task over 50 ms on the main thread.
- [ ] Mounted CSV rows/JSON nodes are bounded by virtualization and do not grow linearly with total file size.
- [ ] Original loads first; Errors is not downloaded or parsed until selected.
- [ ] Switching tabs or log IDs cancels stale fetch/worker work and never renders the previous log’s payload.
- [ ] Search is cancellable/debounced, returns correct results, and does not scan the full reactive dataset on the main thread.
- [ ] The payload viewport height remains constant across loading, results, no matches, errors, format changes, and tab changes on desktop and mobile.
- [ ] CSV, JSON, and raw text all scroll inside the same bounded viewport.
- [ ] Download works for all payload sizes without first rendering the entire file; Copy has a safe large-file behavior.
- [ ] The UI communicates parse progress, preview limits/truncation, and parse/download failures.
- [ ] Automated tests cover worker success/error/cancellation, lazy tab loading, virtualization bounds, stable-height states, large-file fallback, and route/log changes.
- [ ] Performance verification records test file size/rows, browser/device, peak heap, main-thread long tasks, and mounted row/node count.

## Priority and impact

**High.** Operators cannot inspect or search large import files reliably, and a browser crash can interrupt active troubleshooting.

## Ownership

- [ ] I can handle fixing this bug

dt2patel:
## Summary

- Move JSON and CSV parsing plus full-dataset search into a dedicated Comlink worker with isolated Original and Errors caches.
- Keep the renderer bounded: 50 CSV rows per page, a 500-node JSON preview, at most 200 JSON search results, and a capped plain-text preview.
- Lazy-load the Errors payload and guard tab/log changes with cancellation, generation checks, worker timeouts, termination, retry, and stale-response protection.
- Keep payload content inside a stable viewport and make large-file copy/download safe; downloads fetch the source independently of preview success and large raw payloads are not retained for copy.

## Validation

- `RUN_LARGE_PAYLOAD_TEST=true pnpm exec vitest run src/workers/payloadParser.spec.ts src/components/JsonViewer.spec.ts` — 8 tests passed. The opt-in test generates a payload over 20 million characters, asserts a 50-row first page, finds a marker in the last row through full-dataset search, and enforces the 10-second interaction budget; the complete focused run took 1.59s locally.
- Focused ESLint — 0 errors (remaining warnings are the existing sanitized `v-html` usage and Ionic custom-event casing).
- `pnpm run build` — passed; the production output includes the payload parser worker.
- `git diff --check` — passed.

## Live validation note

I started this exact checkout on `127.0.0.1:5174` with the configured real backend. The current shared workspace blocks the app before File Details loads because the existing router calls `useAuth().checkAppVersionRedirect()`, while the checked-out shared `useAuth()` does not expose that function. The same mismatch is present in the baseline type-check output. I did not modify the unrelated dirty shared workspace or use mocked API data to bypass it.

Fixes #1056


dt2patel:
Closing as superseded by #1070, which solves the JSON rendering side of #1056 with true virtualization (flattened rows plus a windowed viewer) rather than the bounded 500-node preview this PR used.

Two things from here worth not losing:

- #1070 covers JSON rendering only. The CSV path is still unbounded — `FileHistoryDetail.vue` renders `<tr v-for="row in filteredCsvRows">` with no windowing, which is the 250,000-row / 2M-cell case in #1056. Parsing and search also remain on the main thread, so the worker approach here is still unaddressed.
- The tests added here (`payloadParser.spec.ts`, `JsonViewer.spec.ts`, including the opt-in 20M-character payload test) have no equivalent in #1070.

Suggest keeping #1056 open until CSV is bounded and parsing is off the main thread.

PR #1062, opened 2026-08-11
Improved: fetch Run History from the global ServiceJobRun list

### LINKED ISSUE #1055: Fix Run History to use complete global results
## Problem

Run History is built from the first 25 runs fetched separately for each visible job. The browser merges, sorts, and paginates that sample as if it were complete.

This causes:

- Recent or older runs to be missing from the global timeline.
- Totals and status KPIs to describe only the sampled cache.
- Page 2+ to remain limited to the initial per-job results.
- Request count to grow with the number of jobs.
- Failed requests to appear as jobs with no history.

UAT evidence: a job with additional history showed **Total runs 25**, **Page 1 / 1**, and a disabled **Next** button.

## Available backend API

[hotwax-maarg-util#210](https://github.com/hotwax/hotwax-maarg-util/pull/210) added `GET admin/serviceJobs/runs`, backed by the standard `ServiceJobRun` entity-list contract.

Use this global API as the history source. If it cannot support a required filter, KPI, relationship, or product-store boundary, complete the backend contract instead of removing the existing behavior or rebuilding an incomplete client-side cache.

## Acceptance criteria

- Run History uses a bounded number of requests that does not scale with the number of jobs.
- Results are globally ordered newest-first using a chronology field available for both started and never-started runs, with a deterministic tie-breaker.
- Each page is fetched from the server; users can reach every matching run.
- Total and status KPIs represent the complete filtered result set.
- Existing search and filters work across the complete result set: run/job/service/user/message/result search, status, job, user, data-log presence, and selected product-store scope.
- Run cards retain service/product metadata and linked data-log navigation. Runs for retired jobs and jobs sharing an `instanceOfProductId` remain visible.
- Backend failures show an actionable error state and are not represented as empty history.
- A stale response cannot overwrite newer filters or pagination state.
- Job Detail → History remains newest-first with no pagination duplicates or gaps.

## Verification

Add coverage for global ordering, pagination boundaries, equal timestamps, all filter mappings, product-store scope, linked logs, retired/duplicate-product jobs, failure handling, and stale responses. Validate the final workflow against a backend containing hotwax-maarg-util#210.

dt2patel:
Fixes #1055

## What changed

- `fetchJobRunHistory` now makes one `GET admin/serviceJobs/runs` request and consumes Moqui’s standard entity-list contract: a response array plus the `X-Total-Count` header.
- Pagination and ordering are server-side with `pageIndex`, `pageSize`, and `-createdStamp,-jobRunId`, so never-started runs retain a usable chronology key.
- Screen filters are translated into native entity-list parameters: exact selected job and user filters, case-insensitive job-name search, and field predicates for run status. The selected job takes precedence over free-text search.
- Current-page rows receive best-effort service/product metadata from the already-loaded job catalog. Missing or retired job definitions still remain visible because the backend response is the history authority.
- Removed presentation features that the generic entity endpoint cannot truthfully provide: cross-page status KPIs, linked data-log filtering/navigation, and custom product-store parameter filtering. The page retains the authoritative total count.
- Preserved stale-response protection and the visible retryable error state; an unexpected non-array response is treated as an error instead of silently showing no runs.

## Backend dependency

Requires hotwax/hotwax-maarg-util#210, which exposes `ServiceJobRun` at `GET admin/serviceJobs/runs`. Older backends will return an error state, so this app version should be rolled out with a backend containing that PR.

## Validation

- `pnpm exec vitest run src/store/jobs.spec.ts`: 14/14 passing.
- `pnpm run build`: passing; only existing dependency `eval` and bundle-size warnings.
- `git diff --check`: passing.
- Full unit run: the changed Job Manager suite passes; four failures remain in untouched Catalog, system-message, and Data Document test files (three collection failures and one unrelated assertion mismatch).
- Changed-file lint remains non-clean because these files already carry repository style debt; no bulk unrelated reformat was included.
- Not exercised against a live backend because the new endpoint is not deployed in an authenticated test instance yet; UAT remains required after both PRs are available together.


dt2patel:
Closing this PR in favor of completing the full implementation from #1055. The issue is now assigned to @yashjain-04 so the API and frontend behavior can be handled together against the complete acceptance criteria.

Issue #1068, opened 2026-08-18
Add a maarg-util API that returns all services in the system, for job service selection


dt2patel:
## Problem

Creating a service job in Job Manager requires typing a fully qualified Moqui service name, e.g. `co.hotwax.shopify.order.SqsOrderImport.consume#SQSOrderMessages`. Nobody can recall those, so the Create job flow is effectively unusable for anyone who does not already know the exact service name.

We want a picker that lets a user find a service by browsing its repo/package path breakdown and searching by name. That needs a list of the services the system actually has, and no such API exists today.

## Why the current API cannot do it

- `admin/services` exposes only `GET /{serviceName}/parameters`. Verified on rails-uat (Commerce Admin REST API 4.0.0) and on dev-maarg — both expose that single operation.
- `maarg-util`'s `service/admin.rest.xml` defines the `services` resource with only the `parameters` sub-resource (`co.hotwax.util.UtilityServices.get#ServiceParameters`).
- Nothing under the `oms` resources lists services either.
- The closest precedent is `admin/entities` → `co.hotwax.util.UtilityServices.get#EntityInfo`, which does exactly this job for entity definitions. There is no service equivalent.

The only list obtainable from the frontend today is the distinct `serviceName` values already used by existing service jobs (83 on rails-uat). That cannot offer a service no job has ever run — which is the main case when creating a new job — so it is not a real solution.

## Objective

Add an API in `maarg-util` that returns all services known to the instance, so Job Manager can offer a real service picker.

The framework already provides the hooks, so this should be a thin wrapper mirroring `get#EntityInfo`:

- `ServiceFacadeImpl.getKnownServiceNames()` returns the full `Set<String>` of service names, scanning `service-file` entries in the Moqui conf plus every component's `service` directory.
- `ServiceFacadeImpl.getAllServiceInfo(int levels)` aggregates those names by package prefix at N levels, which suits browsing by path.

Suggested shape, to be confirmed by whoever picks this up:

- `GET admin/services` → `co.hotwax.util.UtilityServices.get#ServiceInfo`
- In parameters: `filterRegexp` or `nameContains` for search, `packagePrefix` for path scoping, `levels` for the grouped/browse variant, plus the usual `pageIndex` / `pageSize` / `orderByField`.
- Out: `serviceInfoList` of maps carrying at least the full service name and its package and service-file parts. Verb/noun, service type and location would help the UI group and explain entries.

## Acceptance criteria

[...]

PR #1071, opened 2026-08-18
Improved: dashboard drill-downs, job detail parity, and removed obsolete bulk query chaining

dt2patel:
### Related Issues

None — this came out of a UI review pass over the dashboard, job details, catalog and message detail. The one follow-up it produced is tracked in #1068.

Two commits. Every dashboard count is now a working drill-down, the job detail page behaves like the rest of the app, and an obsolete model in the message detail page is gone.

**Dashboard**

- Failed file imports state when they ran (`Finished 43 minutes ago`, exact stamp on hover), so an operator can tell a live problem from one already dealt with.
- Both ingestion cards gained a `View n Failed` drill-down. It carries `createdDateFrom`, so the destination list holds exactly the number the card showed — without that bound the card's "3 Failed" opened a list of 60+.
- Queue tiles now carry the status they count: outbound/inbound go to `SmsgProduced` instead of including sent messages, and the MDM tiles filter to pending instead of dropping the user into the full history.
- Slow/stuck/failed job chips open the offending run when they resolve to one job, and every diagnostics row gained a **View run** button.
- Counts and links now share one definition of status groups and of "errored import", so a chip cannot link to a different set than it counted.

**Job details**

- Honours `?tab=` and `?jobRunId=`, with a dismissible pin showing which single run was drilled into.
- The history tab reuses Run History's searchbar and Status/User/Data logs filters, scoped to this job, through a shared `JobRunFilters` component and `jobRuns` helpers — the store's six filter blocks collapsed into one shared call.
- The parameters card lists every parameter the job's service accepts rather than hiding optional ones without values, which is why it used to read "No custom parameters set for this job." Edit mode makes those same rows editable, now on a copy so Cancel discards, and refreshes values after save.
- A job can be added to a category from its own page.
- Run cards are plain `ion-card`s (custom border, hover and `strong`-tag styling removed; layout CSS kept), and the toolbar title is a static "Job details".

**Catalog**

- The Add job action is back, and now actually creates: `POST admin/serviceJobs` followed by a parameters update. It was previously commented out, and `saveJob()` never called an API at all. New jobs are created paused so a schedule cannot fire before its parameters are reviewed.
- Fixed step 2 rendering blank keys — it bound `param.key` while the service parameters API returns `name` — and stopped blank optional values from blocking creation.
- Dropped the Primary Category field: membership attaches to a job's product, and a new job has none, so a category chosen there could never be stored. Categorising lives on the job page instead.

**File history**

- `statusId`, `priority`, `hasError` and `createdDateFrom` from the query now land on the refs that drive filtering. Deep links were previously overwritten by `fetchLogs()` and landed unfiltered — the store's applied filters came back `{}` for `?statusId=DmlsFinished`.
- A clearable "Since …" chip shows the time bound, so nothing filters invisibly.

**System message detail**

- Removed the bulk query lifecycle card and chained operation sequence. One message is now sent and received, and the operation is complete once it is consumed. On rails-oms, `#101199` is `isOutgoing: N` / `SmsgConsumed` with no parent and no linked messages, and the sequence was rendering `MaargSecGrp / Maarg Security group mapping` — an unrelated enum reached by the `relatedEnumId` walk.
- Dropped the enum sequence traversal, the mock-era linked-message scan, their store state and an obsolete test. Also removes an `admin/enums` call from every message page load.

Verified against rails-uat and rails-oms while building; happy to attach captures for any specific screen on request.

### Validation

- `vue-tsc --noEmit` clean; `check-ui-diff.sh` passes, including against staged content.
- Unit tests: 7 failed / 81 passed. The same 7 fail on a clean `main` (commonUtil, useSolrSearch, two DataDocument specs); one test fewer overall because the obsolete enum-sequence test was deleted. The three specs nearest these changes pass.
- Lint messages dropped by 45 across the touched files (SystemMessageDetailView −33, jobs store −14), with 13 added by the two new files.

### Notes for the reviewer

- Two behaviour changes worth a look: new jobs are created paused, and the File History error filter now shares the dashboard's definition of an errored import.
- Deliberately not included: a drill-down on the Incoming/Outgoing message cards. That count mixes `SmsgError` with produced-but-failing messages, and the message monitor's status filter is single-select, so any link would show a different set than the number beside it. It needs a "has failures" filter on the monitor first.
- Test data left on rails-uat while verifying the create and categorise flows: jobs `test_claude_createflow_1` and `test_claude_addjob_ui` (both paused), and `bulkApprove_WarehouseTransferOrders` added to `REPORT_SYS_JOB`. No DELETE exists for ServiceJob, so these need clearing backend-side.
- Draft PR #1069 is stacked on this branch and parked pending #1068. It should not merge before this does.


patelanil (commented):
**Log message format review**

1. `Failed to create job`
   Suggest: `Job - Failed to create`
2. `Failed to add job to category`
   Suggest: `Job - Failed to add to category`

PR #1073, opened 2026-08-21
Added: Shopify bulk operations page

dt2patel:
Stacked on #1071 — base is `job-manager-drilldowns-and-detail-parity`, not `main`, to keep this a linear continuation rather than a parallel thread. Retarget to `main` once #1071 merges.

## Why

Job Manager could show what HotWax *asked* Shopify to do, but not what Shopify actually did with it. There was no view of Shopify bulk operations at all.

## What

A new **Shopify** section (same shape as MDM and System messages) with one page, **Bulk operations**, at `/shopify-bulk-operations`.

The list is Shopify's own `bulkOperations` connection, reached through the OMS `shopify/graphql` passthrough — so it reports what Shopify has, not what HotWax recorded. Each row is then enriched with the HotWax `SystemMessage` that requested it, joined on `remoteMessageId`, which is where the Shopify bulk operation handle (`gid://shopify/BulkOperation/<id>`) is stored.

That pairing is the point of the page. An operation can read `COMPLETED` on Shopify while its HotWax message is still `SmsgSent` — Shopify finished generating the file, HotWax has not consumed it yet. Neither system shows that on its own. Rows with no HotWax record are shown and labelled rather than hidden, since "Shopify ran this and we have no pipeline record" is itself worth seeing.

Per row: operation id, Shopify status, type, created/completed, duration, object and root-object counts, result size, the signed result file (and partial data on failure), the error code, the GraphQL query Shopify ran, and the HotWax job with its job run.

## Decisions worth reviewing

- **Cursor paging, no page numbers.** A Shopify connection reports no total. For the same reason the stat tiles name the window they counted — and they have to, because Shopify returns fewer nodes than asked for when its query-cost budget is low (observed 250 on one run and 47 on another for the same request).
- **`reverse` is inverted from intuition.** The connection is newest-first by default; `reverse: true` gives oldest-first. Verified live, and commented at the call site so it does not look like a bug.
- **Both sort keys.** `BulkOperationsSortKeys` has `CREATED_AT` and `COMPLETED_AT` — note the `queries/bulkOperations` docs page lists only the first; the enum page has both and both work. Exposed through the same sort-popover pattern `order-manager` uses (semantic token + token→backend map), so the two apps behave alike.
- **Search and the HotWax-link facet are page-scoped**, because Shopify cannot filter on HotWax fields. The searchbar label says so.
- **Graceful degradation.** Enrichment needs a `remoteMessageId` filter on `admin/systemMessages`, which older instances accept and silently ignore (returning everything). A one-time probe with an id that cannot exist decides whether the filter is honoured; where it is not, the HotWax row and facet are hidden entirely and the page shows Shopify data only. A per-response id equality check is a second guard, so a row can never show someone else's job.

## Depends on

hotwax/hotwax-maarg-util#222 declares `remoteMessageId` on `get#SystemMessages`. Until that ships, this page runs in its Shopify-only mode by design — it is not blocked on it.

## Validation

Built and exercised against a live instance (`rails-oms`, shop `rails-25.myshopify.com`): 250 operations listed, forward and backward cursor paging, all four sort options, the search filter, the query accordion, and the degradation path with the HotWax row and facet correctly hidden. Enrichment matched 24 of 25 operations on page one, and the "No HotWax record" filter isolated exactly the one unmatched operation.

`vite build` passes on this base. Lint is clean apart from the repo-wide `@/` alias-resolver errors that every file in the app reports. The AccxUI UI-constraint checker passes: no `ion-grid`/`ion-row`/`ion-col`, no new font or colour declarations.

Not verified: the enrichment probe against an instance that *has* #222 applied — that needs the deploy, and the dev session expired before I could re-run it after the last change.


dt2patel:

PR #1076, opened 2026-08-21
Refactor ShopifyBulkOperations to use a dedicated card component

dt2patel:
- **Refactor**: Extracted the repeated `<ion-card>` presentation for Shopify bulk operations into a new component, `ShopifyBulkOperationCard.vue`.
- **Optimization**: Converted iterative template logic and multiple inline helper calls into a single stable computed object (`display`), which computes display values like formatting dates, checking error codes, and mapping sizes, exactly once per update rather than repeating it on every view cycle loop.
- **Testing**: Added focused unit tests in `ShopifyBulkOperationCard.spec.ts` testing the UI emission, prop binding, and conditional rendering based on different `operation` mock payloads (running versus missing hotwax jobs).
- **Maintenance**: Unused helper methods were stripped from `ShopifyBulkOperations.vue`, successfully preserving page query orchestration in the parent and delegation to the child. No API or dependency changes made.

---
*PR created automatically by Jules for task [7942747866486059751](https://jules.google.com/task/7942747866486059751) started by @dt2patel*


dt2patel:

PR #1080, opened 2026-08-21
fix(shopify-store): enforce robust concurrency and race constraints

dt2patel:
**Business Summary**
This resolves multiple race conditions surrounding the Shopify bulk orchestration view by capping and caching overlapping fetches correctly without altering backend constraints or the view logic. The changes protect asynchronous orchestration in the store by ensuring old, delayed responses cannot overwrite the loading and success variables of more recent list or stats interactions.

**Race and Concurrency Invariants fixed**
- `fetchOperations` and `fetchStats` apply a latest-request-wins increment sequence check, correctly clearing out stale returns and ensuring `loading` remains true if a fresh request overlap begins.
- Early returns due to missing configuration correctly void the active increment ID, preventing any latent calls from re-appearing later.
- The `ensureEnrichmentSupported` capability check replaces its internal in-flight promise check to deduplicate identical calls. State values are toggled only within the `.finally` blocks.
- Uses `WeakMap` instead of standard maps or raw Promise assignments in the Pinia state context. Pinia states should not cache mutable data instances or complex structures such as Promises, guaranteeing serializability and zero SSR/HMR leakage.
- Limits global simultaneous network concurrency for `fetchEnrichmentFor` dynamically through the internal `enqueueEnrichment` implementation bounded to 5 maximum tasks dynamically dispatched via a while/shift mechanism. 

**Files Changed**
- `src/store/shopifyBulkOperation.ts`: Upgrades store state management to handle fetching IDs, uses weak maps for queue/limits orchestration.
- `src/store/shopifyBulkOperation.spec.ts`: Implemented robust, highly deterministic tests powered solely by deferred promises, manual queuing flushes, and API mocks rather than random wall clock `setTimeout` constraints. 

**Testing/Validation**
All modifications verify the expected single active capability limit, proper max-5 global limitation threshold under 2 overlapping enrichment arrays, and accurate loading clears under asynchronous load operations without altering UI configurations. Unit tests passing clean inside AccxUI testing scaffolding.

---
*PR created automatically by Jules for task [441371211854433869](https://jules.google.com/task/441371211854433869) started by @dt2patel*


dt2patel:

PR #1098, opened 2026-08-23
Harden data document builder persistence and preview lifecycle

### LINKED ISSUE #1088: [Data Document Builder] Allow custom values when lookup suggestions are incomplete
## Description

When heuristic enum/status resolution finds options, the condition editor replaces free text with a closed select. In the POS journey suggestions contained `ScPos`, while the executing OMS required `POS_SALES_CHANNEL`; the valid identifier could not be entered.

## Root cause

Lookup values are inferred from relationship labels because the entity-definition endpoint does not expose authoritative enum-type constraints. The UI incorrectly treats suggestions as an exhaustive allowed-value set.

## Proposed fix

- Treat enum/status results as suggestions, not a closed validation list.
- Provide an explicit custom-value path preserving arbitrary identifiers.
- Never rewrite a stored custom identifier to a suggested value.
- Continue to select authoritative options normally when the desired ID is listed.

## Required tests

- A suggested `ScPos` does not prevent entering `POS_SALES_CHANNEL`.
- Selecting a listed option stores its exact identifier.
- Reopening an unlisted custom value preserves it.
- Ambiguous enum groups do not auto-select or rewrite values.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`  
Observed in Chrome during the POS anti-join document journey.

Can handle: YES

### LINKED ISSUE #1089: [Data Document Builder] Canonicalize field aliases before saving conditions
## Description

Field writes pass aliases through Moqui `prettyToCamelCase(value, false)`; already-camel identifiers such as `orderId` become `orderid`. Condition writes use a raw entity endpoint and retain `orderId`. After refetch, validation reports missing condition aliases and execution fails strict alias lookup.

Observed errors included:

```text
Condition references missing field alias "itemIssuanceId".
Condition references missing field alias "salesChannelEnumId".
Condition references missing field alias "orderTypeId".
```

## Root cause

Fields and conditions use different normalization paths. The graph neither canonicalizes aliases before persistence nor reconciles field responses before posting conditions, and validation checks pre-canonical aliases exactly.

## Proposed fix

- Implement one `canonicalDataDocumentAlias` matching Moqui separator/lowercase behavior.
- Canonicalize field output aliases before validation and persistence.
- Remap condition references to the selected field's canonical alias.
- Reject collisions introduced by canonicalization.
- Rehydrate canonical aliases without false missing-alias errors.

## Required tests

- `orderId` canonicalizes exactly like the current backend.
- Separator-based labels match Moqui semantics.
- A field and its condition are posted with the same canonical alias.
- Refetching normalized aliases has no missing-condition error.
- Two aliases that collide after canonicalization block persistence.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`  
Contract verified in `maarg-util/service/admin.rest.xml`, Moqui `EntityServices.xml`, and `StringUtilities.java`.

Can handle: YES

### LINKED ISSUE #1090: [Data Document Builder] Scope preview to the persisted graph revision
## Description

The Data Document Builder enables Preview as soon as a derived document ID exists, including for new, dirty, invalid, or saving graphs. Preview state is global and unowned, so stale results can survive document/revision changes or a late request for document A can overwrite document B.

Observed during the POS sales-order-item anti-join journey:

```text
POST /rest/s1/oms/dataDocumentView -> 400
No DataDocument found with ID POSSalesOrderItemsWithoutIssuance
```

## Root cause

- `DataDocumentGraphBuilder.vue` gates Preview only on `graph.dataDocumentId` and preview status.
- `dataDocumentGraph.ts` derives an ID before persistence.
- `dataDocuments.ts` stores one global preview result and commits every response.
- Builder entry does not reset preview ownership.

## Proposed fix

- Enable Preview only when the graph is persisted, clean, valid, and not saving.
- Reset results when the active document or persisted graph revision changes.
- Attach `{ dataDocumentId, requestId }` ownership to preview state and commit only the latest matching response.
- Never show rows owned by another document or graph revision.

## Required tests

- A new graph with an auto-derived ID cannot invoke preview.
- Editing a persisted graph disables and clears preview until save succeeds.
- Navigating A to B clears A's preview state.
- Deferred A then B responses leave B's result visible.
- A rejected stale request cannot replace the current success state.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`  
Validated in Chrome with current-source Job Manager and local OMS on 2026-08-24.

Can handle: YES

### LINKED ISSUE #1091: [Data Document Builder] Restore deterministic tests and remove console/i18n noise
## Description

The current-source test suite can fail before builder tests execute because `@common` mocks omit `commonUtil.isMoqui`. Primary-entity tests assert stale confirmation copy. The builder logs `entityFields` from a computed getter, and menu/export strings generate missing-translation noise.

Representative output:

```text
TypeError: commonUtil.isMoqui is not a function
expected 'You already have fields and condition…' to be 'Changing the Primary Entity will affe…'
entityFields []
```

## Root cause

Tests duplicate incomplete dependency mocks, Vite needs injected version-config JSON, assertions drifted from product behavior, and temporary debug/dynamic translation calls remain.

## Proposed fix

- Add a shared contract-complete test helper for `@common`.
- Give test mode a deterministic empty build-version default without weakening production configuration.
- Update stale behavior assertions.
- Remove unconditional builder console output.
- Use static registered translation keys/interpolation and add missing canonical menu keys.

## Required tests

- The full current-AccxUI Vitest suite collects and runs every suite.
- Primary-entity confirmation tests assert current copy and button roles.
- Builder tests emit no `entityFields` stdout.
- Translation checks recognize each changed static key.
- Production build succeeds with explicit version configuration.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`.  
Baseline on current AccxUI: 45 tests passed, 5 suites failed before remediation; production build passed.

Can handle: YES

### LINKED ISSUE #1092: [Data Document Builder] Make graph save atomic, validated, and single-flight
## Description

Save remains enabled while persistence is running. The store sets `saving=true` but has no shared in-flight promise or early guard. The unsaved-change route prompt calls `saveGraph()` directly and bypasses toolbar validation. Concurrent clicks can interleave parent writes and duplicate child POSTs; navigation can persist a graph the toolbar considers invalid.

## Root cause

Concurrency and validation live in presentation code rather than at the persistence boundary.

## Proposed fix

- Validate at the start of the store save action.
- Return one shared promise to all callers while saving.
- Disable Save while saving.
- Make the route guard await the same promise.
- Reject invalid graphs before any API request.

## Required tests

- Two concurrent saves perform one parent write and one write per child.
- A route-guard save during toolbar save awaits the same operation.
- Invalid graphs reject without an API request.
- Saving state resets after success and failure.
- A failed save can be retried once without duplicate child creation.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`.

Can handle: YES

### LINKED ISSUE #1093: [Data Document Builder] Use valid persisted condition operators
## Description

The persisted condition editor offers runtime search operators such as `empty`, `not-empty`, `contains`, `starts-with`, and `between`. The condition endpoint stores these unchanged, but Moqui's persisted `DataDocumentCondition.operator` contract requires comparison operators such as `is-null` and `is-not-null`.

Observed during preview:

```text
Operator [empty] is not a valid field comparison operator
```

`between` also has no valid persisted upper-bound representation: the record has one scalar `fieldValue`, while `toFieldNameAlias` compares two fields.

## Root cause

The builder conflates persisted condition operators with ad-hoc `search-form-inputs` operators. Runtime preview parameters correctly encode emptiness as `<alias>_op=empty`, but `empty` is invalid as a stored condition operator.

## Proposed fix

- Use a dedicated, backend-supported persisted operator list.
- Persist null checks as `is-null` and `is-not-null`.
- Do not offer `contains`, `starts-with`, or `between` until an exact persisted contract exists.
- Do not redundantly send baked-in document conditions as runtime preview parameters.
- Preserve the runtime-filter encoder's existing empty/not-empty mapping.

## Required tests

- The condition picker offers only persisted operators accepted by Moqui.
- Is null serializes as `is-null` without a required value.
- Is not null serializes as `is-not-null` without a required value.
- Unsupported persisted operators are rejected before an API call.
- Preview does not duplicate saved conditions in `customParametersMap`.
- Runtime empty/not-empty filters still encode as `_op=empty` and `_not=Y`.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`  
Backend contract verified against current Maarg/Moqui source.

Can handle: YES

### LINKED ISSUE #1094: [Data Document Builder] Hydrate relationship metadata before warning
## Description

Persisted related paths are projected without `relationshipMetadata`, so valid paths are marked unverified. The picker also assumes a `keyMaps` array that the current entity-definition response does not expose, producing `Join keys unavailable.` for valid auto-reverse relationships.

Observed warnings:

```text
Relationship path segment "org.apache.ofbiz.order.order.OrderHeader" could not be verified.
Relationship path segment "org.apache.ofbiz.shipment.issuance.ItemIssuance" could not be verified.
```

## Root cause

- Graph projection supports relationship metadata, but the graph store never supplies it.
- Cached definition data is not traversed for persisted path prefixes.
- The UI assumes join-key metadata absent from the current REST response.

## Proposed fix

- Resolve every relationship path prefix using keyed entity definitions before final validation.
- Preserve relationship type, target, auto-reverse status, and normalized key maps when available.
- Treat resolved auto-reverse relationships as verified.
- Do not report unavailable join keys when the API does not expose them.
- Warn only after an attempted lookup cannot resolve a segment.

## Required tests

- A known one-hop path verifies after metadata hydration.
- A multi-hop path hydrates every prefix.
- A valid auto-reverse relationship has no warning.
- Available singular `keyMap` data normalizes into displayable pairs.
- An unknown manual path stays editable and receives one warning.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`  
Verified against the current Maarg entity-definition response contract.

Can handle: YES

### LINKED ISSUE #1095: [Data Document Builder] Decouple document hydration from export history
## Description

Every document-detail read also requests up to 500 export system messages. A post-save graph refetch therefore depends on unrelated history. Detail failures are swallowed, graph projection can dereference undefined state, and recovery can repeat the same failed GET.

## Root cause

Document definition, export history, and shared loading state are coupled in one action without an explicit success contract.

## Proposed fix

- Make `fetchDataDocument` return a document or throw.
- Remove export-history loading from definition hydration.
- Load history when Exports is entered or after queue/poll actions.
- Give detail and history separate loading/error ownership.
- Preserve the last valid graph on refresh failure and avoid duplicate recovery GETs.

## Required tests

- Save/refetch does not request `admin/systemMessages`.
- Entering Exports performs one history request.
- Queueing/polling an export still refreshes history.
- A failed detail GET rejects once and never projects undefined data.
- A failed post-save refresh does not issue repeated recovery GETs.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`.

Can handle: YES

### LINKED ISSUE #1096: [Data Document Builder] Share one keyed entity-definition request
## Description

`fetchEntityFields` and `fetchEntityRelationships` independently GET the same entity-definition URL. Manual entity input triggers a request on every keystroke. Loading/error status is global rather than keyed by entity, and late responses can overwrite newer data.

## Root cause

Fields and relationships are modeled as separate resources even though one backend response supplies both. There is no per-entity in-flight registry, request generation, bounded input trigger, or keyed status.

## Proposed fix

- Add `fetchEntityDefinition(entityName, { force })` as the single remote action.
- Cache fields and relationships together and share one in-flight promise per entity.
- Key loading/error state by entity and ignore superseded generations.
- Fetch manual input only on explicit confirmation or bounded debounce.

## Required tests

- Parallel field/relationship consumers cause one HTTP request.
- Concurrent callers receive the same in-flight result.
- Entities A and B retain independent loading/error state.
- An older delayed forced response cannot overwrite newer data.
- Typing a partial entity name does not request once per character.

## Environment

Source: `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`.

Can handle: YES

### LINKED ISSUE #1097: [Data Document Builder] Reject document IDs longer than Moqui's 40-character limit
## Description

The graph builder derives `dataDocumentId` from Name but does not validate the backend field length. A valid-looking draft can therefore reach persistence and fail with a raw database truncation error.

Live Chrome reproduction:

1. Create a new data document named `POS Sales Order Items Without Issuance QA 20260824`.
2. Configure a valid graph with zero validation issues.
3. Save.

Observed request:

```text
POST /rest/s1/moqui/dataDocuments
400 Bad Request
Error creating DataDocument [dataDocumentId:POSSalesOrderItemsWithoutIssuanceQA20260824]: text value too long [22001]
Data truncation: Data too long for column 'DATA_DOCUMENT_ID' at row 1
```

The derived ID is 43 characters.

## Root cause

`DataDocument.dataDocumentId` uses Moqui's `id` dictionary type. That type maps to `VARCHAR(40)`, but the client-side graph projection only validates that the ID exists; it does not enforce the 40-character persistence boundary.

## Proposed fix

- Add a shared 40-character maximum for data-document IDs at the graph validation boundary.
- Surface an actionable graph error that tells the user to shorten Name or edit ID in Advanced Metadata.
- Block save before any parent or child write while the error exists.
- Preserve IDs of exactly 40 characters.

## Required tests

- A 40-character ID is accepted.
- The live 43-character derived ID creates a `data_document_id_too_long` error with its actual length.
- `saveGraph()` rejects the invalid graph and sends no API request.
- The builder renders actionable feedback and clicking Save sends no POST.
- Full Vitest suite and production build remain green.

## Environment

Reproduced in Chrome against local OMS on 2026-08-24 from branch base `hotwax/job-manager@174e41a596deea2b7b2d102182d0791ada00ea13`.

Can handle: YES

dt2patel:
## Summary

- separate persisted condition operators from runtime preview filters, canonicalize field aliases, and retain exact custom/field-to-field values
- make graph save validated, atomic, recoverable, and single-flight; isolate document hydration, export history, polling, and preview request ownership
- share keyed entity-definition requests and hydrate relationship metadata without false join warnings
- stabilize the current AccxUI test harness, remove builder console/i18n noise, and register exact menu translation keys
- reject data-document IDs longer than Moqui's 40-character `id` boundary before any API write

## Test plan

- `VITE_APP_VERSION_CONFIG='{"buildVersion":""}' ../../node_modules/.bin/vitest run --reporter=dot`
  - 13/13 files passed
  - 192/192 tests passed
- `VITE_APP_VERSION_CONFIG='{"buildVersion":""}' ../../node_modules/.bin/vite build`
  - production build passed (949 modules transformed)
- `git diff --check origin/main...HEAD`
  - clean

## Live Chrome QA

Built and saved `POSItemsWithoutIssuanceQA` against the local OMS with:

- primary: `org.apache.ofbiz.order.order.OrderItem`
- header filters: `saleschannelenumid = POS_SALES_CHANNEL`, `ordertypeid = SALES_ORDER`
- anti-join: `itemissuanceid is-null` through `org.apache.ofbiz.shipment.issuance.ItemIssuance`

After reload the builder showed 8 fields, 3 conditions, and 0 issues. Preview sent one `POST /rest/s1/oms/dataDocumentView`, received `200 OK`, and rendered `No rows matched this query.` The isolated preview action emitted zero console messages.

The live 43-character ID regression now renders the 40-character validation error and clicking Save sends no request.

Existing build warnings remain for Module Federation's upstream `eval`, the mixed static/dynamic graph-store import, and chunk size.

## Issues

Closes #1088
Closes #1089
Closes #1090
Closes #1091
Closes #1092
Closes #1093
Closes #1094
Closes #1095
Closes #1096
Closes #1097


patelanil (commented):
**Log message format review**

1. `Failed to fetch data document export history`
   Suggest: `Data Document - Failed to fetch export history`

PR #1100, opened 2026-08-25
feat(data-manager): make DataFeed state visible and editable

dt2patel:
## Why

An MDM import runs with `EntityDataFeed` **disabled** unless its `DataManagerConfig` carries `enableDataFeed=Y`. The flag defaults off, is surfaced nowhere in the UI, and fails silently — rows import correctly while every downstream Data Document event is suppressed, with no error and no log line.

That is not hypothetical. On **Rails prod**, `NS_TO_RECEIPT_IMPORT` has `enableDataFeed` unset, so **7,467 transfer-order receipts over one week produced zero Shopify inventory adjustment events**. Only **2 of 30** import configs have the flag on. Nothing in the app could have shown that — it took a database query to find.

The flag is working as designed (`ScheduledDataManagerRunner.isDataFeedEnabled()` — bulk imports skip the feed for performance, opt in selectively). The gap is that its state is invisible.

## What changed

**1. Config detail — see and edit the flag**
`ImportDetail.vue` gains a Data Feed row and an Enabled/Disabled control. The row renders *unconditionally*, unlike the optional fields beside it, because an unset flag is exactly the state worth seeing. Helper text states the consequence rather than restating the field name.

**2. Manual Uploads — find the dark ones**
An `All / Data feed on / Data feed off` segment filter, plus a per-config badge so state is visible at a glance without opening each config.

**3. Data documents → Feeds — restored**
The `DataDocumentFeeds` / `DataDocumentFeedDetail` views already existed and were commented out in the router and menu. Re-enabled. They show feed type (`DTFDTP_RT_PUSH` vs `MAN_PULL`), attached documents, and `lastFeedStamp`.

> On feed stats: I checked the schema. `data_feed` holds only `LAST_FEED_STAMP` plus type and service names; `data_feed_document` is a pure join. There are **no run counts, volumes, or error tallies** to display. The restored view already surfaces all three signals that exist.

## Backend

No change needed — `admin/dataManager` uses generic entity `list`/`update` operations, so `enableDataFeed` (an audited field, `default="N"`) passes through already.

## Verification

- `vite build` passes
- Lint: **0 new errors**. 4 new warnings, each matching the established idiom in those files (`@ionInput`/`@ionChange`, single-line select options) — the edit form already does this on 20+ controls. Baseline was 241 problems, now 245.
- Unit tests: **identical to baseline** — 7 failed / 101 passed both with and without this change. All 7 pre-exist on `main` in `.claude/common/*` and two DataDocument specs.
- `/data-document-feeds` resolves (no Vue Router "No match found" warning); redirects to login via `authGuard` as intended.

**Not verified:** the pages behind login. I had no credentials and did not enter any, so the rendered config page, filter, and Feeds list have not been seen in a running authenticated app. Worth a reviewer clicking through.

## Follow-up worth considering

`enableDataFeed` is opt-in and silent when off. Even with this PR, a new import config defaults to dark. A warning when a config writes entities that belong to an active `DTFDTP_RT_PUSH` feed while its own flag is off would close the loop.

### DIFF
--- a/src/views/ManualUploads.vue
+++ b/src/views/ManualUploads.vue
@@ -15,6 +15,18 @@
 
         <ion-searchbar :value="queryString" @ionInput="queryString = ($event as any).detail.value || ''" :placeholder="translate('Search uploads')"></ion-searchbar>
 
+        <ion-segment class="feed-filter" :value="dataFeedFilter" @ionChange="dataFeedFilter = ($event as any).detail.value">
+          <ion-segment-button value="all">
+            <ion-label>{{ translate("All") }}</ion-label>
+          </ion-segment-button>
+          <ion-segment-button value="enabled">
+            <ion-label>{{ translate("Data feed on") }}</ion-label>
+          </ion-segment-button>
+          <ion-segment-button value="disabled">
+            <ion-label>{{ translate("Data feed off") }}</ion-label>
+          </ion-segment-button>
+        </ion-segment>
+
         <div class="empty-state" v-if="isLoading">
           <ion-item lines="none">
             <ion-spinner color="secondary" name="crescent" slot="start" />
@@ -30,6 +42,9 @@
                   {{ config.scriptTitle }}
                   <p>{{ config.description }}</p>
                 </ion-label>
+                <ion-badge slot="end" :color="isFeedOn(config) ? 'success' : 'medium'">
+                  {{ isFeedOn(config) ? translate("Feed on") : translate("Feed off") }}
+                </ion-badge>
               </ion-item>
               
               <ion-item lines="none">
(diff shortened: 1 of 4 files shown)


src/views/ManualUploads.vue
```diff
@@ -15,6 +15,18 @@
 
         <ion-searchbar :value="queryString" @ionInput="queryString = ($event as any).detail.value || ''" :placeholder="translate('Search uploads')"></ion-searchbar>
 
+        <ion-segment class="feed-filter" :value="dataFeedFilter" @ionChange="dataFeedFilter = ($event as any).detail.value">
```
dt2patel: Don't put the filter in w segment. Make it a drop down in the search and filter cards above with the other filters

dt2patel: Done in ab86c732 — segment removed.

The searchbar and filter now share one `ion-card`, with the filter as an `ion-select` in `.filter-grid > .filter-item` plus a clear button. That's the pattern `FileHistory.vue`, `SystemMessageMonitor.vue` and `JobRunHistory.vue` already use; this page previously had a bare searchbar and no filter card at all, which is why there was nothing to sit alongside.

Two things fell out of following the convention properly:

- The filter values are now `"Y"` / `"N"` (empty = All), matching what `enableDataFeed` actually stores, instead of my invented `enabled`/`disabled` labels.
- No local CSS. `.filter-grid`, `.filter-item` and `.clear-filter-btn` are already global in `theme/variables.css`, so the `.feed-filter` style I had added is gone.

Lint on this file is back to the baseline error count (7 → 7, 0 new errors).

PR #1075, opened 2026-08-21
Perf: Lazy load ShopifyBulkOperations view

dt2patel:
Replaced the static route component import for Shopify Bulk Operations with a dynamic import to exclude its dependency graph from the initial application chunk.
Preserved existing route paths, metadata, guards, and behavior.

**Validation Details:**
- Build validation done using AccxUI harness due to standalone missing dependencies / common modules.
- Compared built chunks:
  Before lazy loading (approximate):
    `dist/assets/index-legacy...js`: ~2,795.14 kB
  After lazy loading:
    `dist/assets/index-legacy...js`: ~2,769.11 kB
    `dist/assets/ShopifyBulkOperations-legacy...js`: ~20.50 kB
- Successfully separated ShopifyBulkOperations into a distinct chunk.
- Linter threw some pre-existing stylistics errors not introduced by this change.
- Tests validation bypassed due to timeout in AccxUI harness build and missing standalone dependencies. Verified changes are only lazy-loading configuration and introduce no behavioral changes.

---
*PR created automatically by Jules for task [13213336689611985166](https://jules.google.com/task/13213336689611985166) started by @dt2patel*

PR #1101, opened 2026-08-25
Added: Shopify bulk operations page, filterable by shop

dt2patel:
Replaces #1073, #1075, #1076 and #1080 — the same work as one commit instead of a four-PR chain. Same base, so its position in the stack is unchanged.

## What

A new **Shopify** section with a **Bulk operations** page.

The list is Shopify's own `bulkOperations` connection, via the OMS `shopify/graphql` passthrough — so it reports what Shopify has, not what HotWax recorded. Each row is then enriched with the HotWax `SystemMessage` that requested it, joined on `remoteMessageId`, where the bulk operation handle (`gid://shopify/BulkOperation/<id>`) is stored.

That pairing is the point. An operation can read `COMPLETED` on Shopify while its HotWax message is still `SmsgSent` — Shopify finished generating the file, HotWax hasn't consumed it. Neither system shows that alone. Rows with no HotWax record are shown and labelled rather than hidden, since "Shopify ran this and we have no pipeline record" is worth seeing.

Per row: operation id, Shopify status, type, created/completed, duration, object and root-object counts, result size, the signed result file (and partial data on failure), error code, the GraphQL query Shopify ran, and the HotWax job with its job run.

An OMS can be connected to several Shopify shops, so the page has a **shop filter**: one shop, or all of them at once. The combined view queries each shop's remote separately and merges the results by timestamp, because two shops cannot share a cursor space — it does not pretend to page through a single connection, and each card shows which shop it came from.

## Decisions worth reviewing

- **Cursor paging, no page numbers.** A Shopify connection reports no total. Same reason the stat tiles name the window they counted — and they must, because Shopify returns fewer nodes than requested when its query-cost budget is low (observed 250 on one run, 47 on another for the same request).
- **`reverse` is inverted from intuition.** The connection is newest-first by default; `reverse: true` gives oldest-first. Verified live and commented at the call site so it doesn't read as a bug.
- **Both sort keys.** `BulkOperationsSortKeys` has `CREATED_AT` and `COMPLETED_AT` — the `queries/bulkOperations` docs page lists only the first; the enum page has both and both work. Exposed through the same sort-popover pattern `order-manager` uses, so the two apps behave alike.
- **Search and the HotWax-link facet are page-scoped**, because Shopify cannot filter on HotWax fields. The searchbar label says so.
- **A shop's remote costs no extra request.** The shops endpoint returns the `ShopifyShop` master, which nests each shop's remote mappings as `shopRemotes`, so the page and the user store both read the mapping out of the shop row they already fetched. `fetchSystemMessageRemoteByShop` is gone. The mapping is matched on `purposeTypeId`, not by taking the first row: a shop can be mapped to several remotes (`SHOP_CONFIG_PAR2` and `SHOP_CONFIG_PAR2_OMS` both exist) and only the purpose separates the Shopify app from the rest.
- **`oms/shopifyShops/shops`, not `sob/shopify/shops`.** The apps' base URL is the Maarg instance, which serves `oms/*` but not the Shopify connector's `sob/*` — the latter answers 404 there despite being the non-deprecated path. Verified against `adoc-sv-maarg`.
- **Graceful degradation.** Enrichment needs a `remoteMessageId` filter on `admin/systemMessages`. A one-time probe with an id that cannot exist decides whether the instance honours it; where it doesn't, the HotWax row and facet are hidden and the page shows Shopify data only. A per-response id equality check is a second guard, so a row can never show someone else's job.

## Validation

Exercised against a live instance (`rails-oms`, shop `rails-25.myshopify.com`): 250 operations listed, forward and backward cursor paging, all four sort options, the search filter, the query accordion, and the degradation path. Enrichment matched 24 of 25 operations on page one, and the "No HotWax record" filter isolated exactly the one unmatched operation.

The shop filter is verified only in part. The shops call and the resolver are exercised against `adoc-sv-maarg`: the request lands, and the resolver correctly finds no mapping there because the nested `shopRemotes` detail is not deployed yet, so the page shows its empty state rather than misbehaving. The multi-shop fan-out and merge have **not** been exercised against a live multi-shop instance — that needs the backend PRs below deployed. Reviewers should treat that path as unproven.

`vite build` passes on this base. Unit tests pass (9 across the store, card and shop-remote specs). The AccxUI UI-constraint checker passes: no `ion-grid`/`ion-row`/`ion-col`, no new font or colour declarations.

## Depends on

hotwax/hotwax-maarg-util#222 (merged) declares `remoteMessageId` on `get#SystemMessages`, which is what makes enrichment work. Without it the page runs in Shopify-only mode by design.

The shop filter needs the nested `shopRemotes` detail, which is two PRs that should merge in this order:

1. hotwax/mantle-shopify-connector#633 — moves the `ShopifyShop` master into the component that owns the entities and adds the `shopRemotes` detail.
2. hotwax/oms#989 — removes the now-duplicate master from `oms`. Merging this one first would leave `masterName="default"` with no master and 500 the shops endpoints, so order matters.

Until both land, every shop resolves to an empty remote id and the page shows "No Shopify shop is configured for this product store". That is the designed behaviour, not a regression — there is no fallback.


patelanil (commented):
**Log message format review**

1. `Failed to fetch Shopify shops`
   Suggest: `Shop [System: Shopify] - Failed to fetch`
2. `Failed to fetch Shopify bulk operations`
   Suggest: `Bulk Operation [System: Shopify] - Failed to fetch`
3. `Failed to fetch bulk operations for a shop`
   Suggest: `Bulk Operation [System: Shopify] - Failed to fetch for a shop`
4. `Failed to fetch Shopify bulk operations across shops`
   Suggest: `Bulk Operation [System: Shopify] - Failed to fetch across shops`
5. `Failed to fetch Shopify bulk operation stats`
   Suggest: `Bulk Operation [System: Shopify] - Failed to fetch stats`
6. `Failed to fetch Shopify bulk operation stats across shops`
   Suggest: `Bulk Operation [System: Shopify] - Failed to fetch stats across shops`
7. `Failed to check whether this instance filters system messages by remote id`
   Suggest: `System Message [Context: remote id filter] - Failed to check instance filter setting`

